PR #2, opened 2023-02-14
add Auth and Interactive

acharb:
Adds the Auth and Interactive functionality for sep24. Mimics the steps the kotlin-sdk is using.

note: Didn't add allowing custom config and httpClient yet, will do in subsequent PR. Default using axios for now.


src/walletSdk/interactive/index.ts
```diff
@@ -0,0 +1,97 @@
...
+      accountAddress,
+      assetCode,
+      authToken,
+      "withdraw",
+      extraFields,
+      fundsAccountAddress
+    );
+  }
+
+  async flow(
```
acharb, line 53: 👋 @quietbits @Ifropc the kotlin sdk has an `assetGet` param [here](https://github.com/stellar/kotlin-wallet-sdk/blob/324ccb6dbcac3c723f8164c6eca0eada6b5d0f58/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt#L127)

why was this needed? I didn't see it used in an example. Any reason we can't just add the code like below (line 72) that automatically throws an error if the asset is not included in the anchor's serviceInfo?

```
   if (!assets.includes(assetCode)) {
      throw new AssetNotSupportedError(type, assetCode);
    }
```

Ifropc: Yes, it's more or less the same (with assetGet it's just a bit less code)

src/walletSdk/interactive/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    assetCode: string,
+    authToken: string,
+    extraFields?: ExtraFields,
+    fundsAccountAddress?: string
+  ) {
+    return await this.flow(
+      accountAddress,
+      assetCode,
+      authToken,
+      "deposit",
```
piyalbasu, line 30: might be nice to have an enum for all possible `type`

piyalbasu (approved):

PR #4, opened 2023-03-27
small fixes found while testing

acharb:


src/index.ts
```diff
@@ -1,4 +1,5 @@
 import * as walletSdk from "./walletSdk";
+import { Keypair } from "stellar-sdk";
 
-export { walletSdk };
+export { walletSdk, Keypair };
```
acharb, line 5: this was an interesting bug. For some reason if the parent project using the sdk creates the keypair, an error is thrown when signing the txn (even when stellar-sdk versions match across projects): `Error: XDR Write Error: [object Object] is not a DecoratedSignature`

I'm leaving as a TODO for now and exporting the Keypair from wallet-sdk to be used instead which fixes the bug, open to discussion though

piyalbasu: Oh interesting - this sounds familiar. I can look into this with you when I'm back in office

PR #5, opened 2023-05-09
add root wallet class

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-786?atlOrigin=eyJpIjoiODFiYTY4ZjUwNTk0NGQwNWEyZDc2NTgyMzQwY2ZlZTYiLCJwIjoiaiJ9)

adds the root wallet class

### DIFF
--- a/src/walletSdk/Anchor/index.ts
+++ b/src/walletSdk/Anchor/index.ts
@@ -5,15 +5,16 @@ import { Auth } from "../auth";
 import { Interactive } from "../interactive";
 import { TomlInfo, parseToml } from "../toml";
 import { ServerRequestFailedError } from "../exception";
-import { Config } from "../";
 
 export class Anchor {
   private homeDomain = "";
   private httpClient = null;
+  private cfg;
 
-  constructor(cfg: Config, homeDomain: string, httpClient) {
+  constructor(cfg, homeDomain: string, httpClient) {
     this.homeDomain = homeDomain;
     this.httpClient = httpClient;
+    this.cfg = cfg;
   }
 
   async getInfo(): Promise<TomlInfo> {
--- a/src/walletSdk/Anchor/index.ts
+++ b/src/walletSdk/Anchor/index.ts
@@ -5,12 +5,15 @@ import { Auth } from "../auth";
 import { Interactive } from "../interactive";
 import { TomlInfo, parseToml } from "../toml";
 import { ServerRequestFailedError } from "../exception";
+import { Config } from "../";
 
 export class Anchor {
   private homeDomain = "";
+  private httpClient = null;
 
-  constructor(homeDomain) {
+  constructor(cfg: Config, homeDomain: string, httpClient) {
     this.homeDomain = homeDomain;
+    this.httpClient = httpClient;
   }
 
   async getInfo(): Promise<TomlInfo> {
@@ -32,6 +35,7 @@ export class Anchor {
     const transferServerEndpoint = toml.transferServerSep24;
 
     try {
+      // TODO - use httpClient
       const resp = await axios.get(`${transferServerEndpoint}/info`, {
         headers: {
           "Content-Type": "application/json",
--- a/src/walletSdk/Anchor/index.ts
+++ b/src/walletSdk/Anchor/index.ts
@@ -6,6 +6,7 @@ import { Interactive } from "../interactive";
 import { TomlInfo, parseToml } from "../toml";
 import { ServerRequestFailedError } from "../exception";
 
+// Do not create this object directly, use the Wallet class.
 export class Anchor {
   private homeDomain = "";
   private httpClient = null;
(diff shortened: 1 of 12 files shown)


src/walletSdk/index.ts
```diff
@@ -1,18 +1,115 @@
...
+  constructor(
+    stellarConfiguration: StellarConfiguration,
+    applicationConfiguration: ApplicationConfiguration = new ApplicationConfiguration()
+  ) {
+    this.cfg = new Config(stellarConfiguration, applicationConfiguration);
+  }
+
+  anchor(homeDomain: string, httpClientConfig: ClientConfigFn = null) {
+    const url =
+      homeDomain.indexOf("://") !== -1 ? homeDomain : `https://${homeDomain}`;
```
Ifropc, line 45: With this approach there should be a configuration option in client to enable http somehow. 
In Kotlin, with one config scheme is changed to http (and I'm setting https by default in URL)
TS should has something like this too

src/walletSdk/Anchor/index.ts
```diff
@@ -5,12 +5,15 @@ import { Auth } from "../auth";
...
 import { TomlInfo, parseToml } from "../toml";
 import { ServerRequestFailedError } from "../exception";
+import { Config } from "../";
 
 export class Anchor {
   private homeDomain = "";
+  private httpClient = null;
 
-  constructor(homeDomain) {
+  constructor(cfg: Config, homeDomain: string, httpClient) {
```
Ifropc, line 14: I'm not an expert in javascript, if class being exported, does it mean constructor is exported too? 
I saw in js SDK all "internal" classes have comments to not create them directly. We should add some comments like that later to internal constructors too.

src/walletSdk/index.ts
```diff
@@ -1,18 +1,115 @@
...
+
+// TODO - https://stellarorg.atlassian.net/browse/WAL-789?atlOrigin=eyJpIjoiZjcwNzZhOTJmNjE1NGRhNTk1NDlkOTExMTYxMDJkZmYiLCJwIjoiaiJ9
+interface ClientConfigFn {}
+interface ClientConfig {}
+class HttpClient {}
 
-export enum NETWORKS {
-  PUBLIC = "PUBLIC",
-  TESTNET = "TESTNET",
+export class Config {
```
Ifropc, line 15: This is an internal class in kt, not sure if it should be exported

src/walletSdk/index.ts
```diff
@@ -1,18 +1,115 @@
...
+  recover(servers, httpClientConfig) {
+    return new Recovery(
+      this.cfg,
+      this.stellar(),
+      this.getClient(httpClientConfig),
+      servers
+    );
+  }
+
+  close() {
```
Ifropc, line 66: Do you need to close clients explicitly in JS?

piyalbasu: we shouldn;t need to explicitly close the httpclient for the JS version

src/walletSdk/Anchor/index.ts
piyalbasu: Yes, the constructor will be exported, too

@Ifropc Can you please link one of the comments in js sdk you're referring to?

src/walletSdk/index.ts
piyalbasu: +1

src/walletSdk/Anchor/index.ts
Ifropc: https://github.com/stellar/js-stellar-sdk/blob/master/src/assets_call_builder.ts#L7

piyalbasu: Gotcha - was just making sure I understood. Yeah, this makes sense. We should include similar comments in our codebase

src/walletSdk/index.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/5/commits/fa22147df8a55a2e01b84faeb3bf578819a0b0ef

src/walletSdk/index.ts
acharb: agreed, removed https://github.com/stellar/typescript-wallet-sdk/pull/5/commits/f6b2de78f41959406599c3bcf7dc9a9e53c02f72

src/walletSdk/Anchor/index.ts
acharb: 👍  https://github.com/stellar/typescript-wallet-sdk/pull/5/commits/48b0bc4c95c87def43143a725f34ae15cc551b67

src/walletSdk/index.ts
acharb: gotchya, I can add that as part of [this ticket ](https://stellarorg.atlassian.net/browse/WAL-789?atlOrigin=eyJpIjoiYjFkOTU5ZDFkZDkxNDZhYmJhNmNiM2U3MTYxZWQyMDQiLCJwIjoiaiJ9)when adding the httpclient stuff. I made a note of it on that ticket

piyalbasu (approved):

PR #7, opened 2023-05-18
Upgrade deps for node v18

piyalbasu:


package.json
```diff
@@ -1,6 +1,9 @@
 {
-  "name": "typescript-wallet-sdk",
+  "name": "@stellar/typescript-wallet-sdk",
```
piyalbasu, line 2: change package name to be consistent with other stellar libs

package.json
```diff
@@ -16,16 +19,21 @@
...
     "stellar-sdk": "^10.4.1",
-    "ts-loader": "8.0.7",
-    "webpack": "4.41.5",
-    "webpack-cli": "3.3.10"
+    "stream-http": "^3.2.0",
+    "ts-loader": "^9.4.2",
+    "url": "^0.11.0",
+    "util": "^0.12.5",
+    "webpack": "^5.83.1",
+    "webpack-cli": "^5.1.1"
```
piyalbasu, line 33: it appears the main issue was that we needed to upgrade webpack v5, which fixes some security vulnerabilities. Because of this, we will need to mandate that users use node >=18

stream-http, url, util, and https no longer come prepackaged with webpack as of v5, so we need to add those manually.

I also upgraded Typescript because why not?

package.json
```diff
@@ -16,16 +19,21 @@
...
   },
   "scripts": {
+    "prepack": "yarn build",
```
piyalbasu, line 36: this script will be needed when we publish to npm; `prepare` tells npm what script to run to pack everything up before publishing to npm when you run `npm publish`

src/walletSdk/index.ts
```diff
@@ -1,7 +1,7 @@
 import StellarSdk, { Networks, Server } from "stellar-sdk";
 
-import { Anchor } from "./anchor";
-import { WalletSigner, DefaultSigner } from "./auth/WalletSigner";
+import { Anchor } from "./Anchor";
+import { WalletSigner, DefaultSigner } from "./Auth/WalletSigner";
```
piyalbasu, line 4: typescript 5 was complaining about improper casing

webpack.config.js
```diff
@@ -15,6 +15,12 @@ module.exports = {
   },
   resolve: {
     extensions: [".js", ".json", ".ts"],
+    fallback: {
+      http: require.resolve("stream-http"),
+      https: require.resolve("https-browserify"),
+      url: require.resolve("url"),
+      util: require.resolve("util"),
```
piyalbasu, line 22: including the polyfills that don't come prepackaged with Webpack anymore

package.json
CassioMG: Awesome, thanks for doing this!

package.json
CassioMG: @piyalbasu just double checking should it be `prepack` or `prepare` as you mentioned?

CassioMG (approved):

package.json
CassioMG: should we update README instructions with `@stellar/typescript-wallet-sdk` as well?

package.json
Shaptic: You can move a ton of these into devDependencies!

acharb (approved):
lgtm!

package.json
```diff
@@ -16,16 +19,21 @@
     "jest": "^29.4.1",
     "ts-jest": "^29.0.5",
     "tslib": "^2.5.0",
-    "typescript": "4.2.3"
+    "typescript": "^5.0.4"
   },
   "dependencies": {
```
Shaptic, line 24: polyfills, webpack, and typescript stuff are all devdeps!

piyalbasu: Ah good catch on webpack. I would argue the polyfills are regular deps as they're being used at runtime and not just used for compilation

piyalbasu:
I forgot to enable commit signing. Here's the new PR: https://github.com/stellar/typescript-wallet-sdk/pull/9

PR #8, opened 2023-05-21
add generic http request maker

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-789)

using axios as the sdk http client (js sdk is also using)


src/walletSdk/index.ts
```diff
@@ -97,13 +100,13 @@ export class StellarConfiguration {
...
-  constructor(
-    defaultSigner: WalletSigner = DefaultSigner,
-    defaultCliengConfig: ClientConfig = {}
-  ) {
+  defaultClient: AxiosInstance;
+  constructor(defaultSigner: WalletSigner = DefaultSigner) {
     this.defaultSigner = defaultSigner;
-    // TODO - default client
-    this.defaultClient = new HttpClient();
+    this.defaultClient = axios.create({
```
acharb, line 106: @Ifropc regarding your comment from https://github.com/stellar/typescript-wallet-sdk/pull/5#discussion_r1189200628

using axios which allows http by default

CassioMG (approved):

src/walletSdk/index.ts
Ifropc: Looking good, thanks!
About the comment above, the idea is that when developer creates anchor like this
`.anchor(my-domain.com)` it defaults to `https://my-domain.com` internally.
However, we want to be able to also use `http://my-domain.com`. In kt wsdk, developer can either create anchor implicitly via `.anchor(http://my-domain.com)` or simply setting default schema to http.
Can we add test, that simple stuff with http schema works? (Not sure if demo wallet allows http schema, but we should at least check that we are making requests to http in underlying classes, such as Auth)
So, simply testing Auth with http should be good. 
(We can merge this PR and create a test in a separate PR)

Ifropc (approved):

src/walletSdk/index.ts
acharb: ah I see what you're saying. I added this test for calling w/ http, lmk what you think (it was quick so included in this PR :) )
https://github.com/stellar/typescript-wallet-sdk/pull/8/commits/a5a817ca7d5d9e49a86fd56100700bfd8bd9ed29

Ifropc: LGTM

src/walletSdk/Anchor/index.ts
```diff
@@ -128,15 +133,18 @@ export class Anchor {
 
     try {
       // TODO - use httpClient
```
Ifropc, line 135: Should we use client here?

src/walletSdk/Anchor/index.ts
```diff
@@ -179,15 +187,20 @@ export class Anchor {
...
-      const resp = await axios.get(`${transferServerEndpoint}/transactions?${queryString.stringify(apiParams)}`, {
-        headers: {
-          Authorization: `Bearer ${authToken}`,
-        },
-      });
+      const resp = await axios.get(
```
Ifropc, line 190: And here

src/walletSdk/Anchor/index.ts
acharb: yup good catch https://github.com/stellar/typescript-wallet-sdk/pull/8/commits/d657eea6d56447f306ef5d2dd90e0bd3b815c7cc

src/walletSdk/Anchor/index.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/8/commits/d657eea6d56447f306ef5d2dd90e0bd3b815c7cc

PR #10, opened 2023-05-23
Add getTransactionBy function

CassioMG:
[[TS] Port getTransactionBy function from Kotlin SDK](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-835)

Note: I did not worry much about typing the transaction object as we have a [dedicated task](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-790) for it

### DIFF
--- a/src/walletSdk/Anchor/index.ts
+++ b/src/walletSdk/Anchor/index.ts
@@ -1,10 +1,11 @@
 import { StellarTomlResolver } from "stellar-sdk";
 import axios from "axios";
+import queryString from "query-string";
 
 import { Auth } from "../auth";
 import { Interactive } from "../interactive";
 import { TomlInfo, parseToml } from "../toml";
-import { ServerRequestFailedError } from "../exception";
+import { MissingTransactionIdError, ServerRequestFailedError, InvalidTransactionResponseError } from "../exception";
 
 // Do not create this object directly, use the Wallet class.
 export class Anchor {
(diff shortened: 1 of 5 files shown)


piyalbasu (approved):
lgtm! I added @Ifropc as a reviewer, as well

src/walletSdk/Anchor/index.ts
```diff
@@ -19,21 +43,23 @@ export class Anchor {
...
+    const stellarToml = await StellarTomlResolver.resolve(this.homeDomain);
+    const parsedToml = parseToml(stellarToml);
+    this.toml = parsedToml;
+    return parsedToml;
   }
 
   async auth() {
-    const tomlInfo = await this.getInfo();
-    return new Auth(tomlInfo.webAuthEndpoint);
+    const toml = this.toml || await this.getInfo();
```
Ifropc, line 53: [nit]
Maybe move this logic to getInfo?

Ifropc (approved):

src/walletSdk/Anchor/index.ts
CassioMG: @Ifropc agree on that

I was actually on the fence on moving this logic to getInfo to make sure getInfo would always fetch fresh TOML values.
But as Anchors should rarely update their TOML having it cached probably won't be an issue in favor of wallet performance. At any rate I think we could have the best of both worlds by adding a boolean input to the getInfo function so that devs could force refresh the TOML whenever needed. Something like this:
```
  async getInfo(shouldRefresh?: boolean): Promise<TomlInfo> {
    if(this.toml && !shouldRefresh) {
      return this.toml;
    }

    const stellarToml = await StellarTomlResolver.resolve(this.homeDomain);
    const parsedToml = parseToml(stellarToml);
    this.toml = parsedToml;
    return parsedToml;
  }
```

What do you think?

CassioMG: Just updated PR with above change

PR #12, opened 2023-05-26
Add watchAllTransactions functionality

CassioMG:
Related task: [[TS] Functionality for watching transactions updates](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-836)

Another PR should be issued for a similar `watchOneTransaction` functionality

Please note it's using [Add getTransactionsForAsset function](https://github.com/stellar/typescript-wallet-sdk/pull/11) PR's branch as `base branch`


src/walletSdk/Anchor/index.ts
```diff
@@ -195,6 +218,183 @@ export class Anchor {
...
+      onMessage,
+      onError,
+      watchlist = [],
+      timeout = 5000,
+      isRetry = false,
+      ...otherParams
+    } = params;
+    
+    // make an object map out of watchlist
+    const watchlistMap: any = watchlist.reduce(
```
acharb, line 257: nit: I think we can remove the `any` here and have typescript infer it?
```suggestion
    const watchlistMap = watchlist.reduce(
```

src/walletSdk/Anchor/index.ts
```diff
@@ -195,6 +218,183 @@ export class Anchor {
...
+    if(!this._transactionsIgnoredRegistry[assetCode]) {
+      this._transactionsIgnoredRegistry[assetCode] = {};
+    }
+
+    // if it's a first run, drop it in the registry for the given asset code
+    if (!isRetry) {
+      this._watchAllTransactionsRegistry[assetCode] = true;
+    }
+
+    this.getTransactionsForAsset({ authToken, assetCode, ...(otherParams || {}) })
```
acharb, line 276: we can just have it as just this right?
```suggestion
    this.getTransactionsForAsset({ authToken, assetCode, ...otherParams })
```

src/walletSdk/Anchor/index.ts
```diff
@@ -195,6 +218,183 @@ export class Anchor {
...
+      .then((transactions: any[]) => { // TOOD - replace with Transaction[] type
```
acharb, line 277: instead let's use async/await to be consistent with other code. Or is there a reason this needs to be .then instead?
```suggestion
      .then((transactions: any[]) => { // TOOD - replace with Transaction[] type
```

acharb (approved):
looks great! I had some smaller questions/nits, not blocking so approving

src/walletSdk/Anchor/index.ts
CassioMG: Nice catch! I thought `otherParams` could become `undefined` in case no optional params were provided but I just discovered it would actually become `{}` in that case :)

src/walletSdk/Anchor/index.ts
CassioMG: @acharb Good point.

Since this is a watcher functionality we have the option to call it `without awaiting anything`, like this:
```
      anchor.watchAllTransactions({
        authToken, 
        assetCode: "SRT",
        lang: "en-US",
        onMessage,
        onError,
      });
```
because we are only interested in the async values returned by `onMessage` and `onError` callbacks whenever they have something

But, in case we add any `await` inside the upper scope of the function body then `watchAllTransactions` would become an `async function` and devs would end up calling it like this:
```
      await anchor.watchAllTransactions({
        authToken, 
        assetCode: "SRT",
        lang: "en-US",
        onMessage,
        onError,
      });
```
^ in this case, the code would await for the `getTransactionsForAsset` API call to complete to only then proceed to the next line which would reduce performance a little bit

I think both ways could work, it's just that not making it an async function seems a bit more convenient and improves performance

How does that sound to you?

acharb: hmm I see your point. I guess this way we're enforcing the watching code to happen async, where as if we make `watchAllTransactions` an `async func` then it would give devs the option to use async or not (by including `await` or not), but may confuse them to always think they need to await.

I don't feel strongly about changing, sgtm to keep!

src/walletSdk/Anchor/index.ts
```diff
@@ -201,6 +224,183 @@ export class Anchor {
...
+  *
+  * On initial load, it'll return ALL pending transactions via onMessage.
+  * Subsequent messages will be any one of these events:
+  *  * Any new transaction appears
+  *  * Any of the initial pending transactions change any state
+  *
+  * You may also provide an array of transaction ids, `watchlist`, and this
+  * watcher will always react to transactions whose ids are in the watchlist.
+  */
+  watchAllTransactions(params: {
```
Ifropc, line 239: Can we move implementation to a separate class? And here just make a delegate call (anchor class is getting quite big)
(Maybe even make sense to move the method too, so it will be `anchor.watcher.watchAllTransaction()`
If we plan to add more functionality to the watcher later.)
Also, if we have a watcher class, they will be able to create multiple watchers

src/walletSdk/Anchor/index.ts
```diff
@@ -201,6 +224,183 @@ export class Anchor {
...
+    authToken: string;
+    assetCode: string;
+    onMessage: (transaction) => void;
+    onError: (error: any) => void;
+    watchlist?: string[];
```
Ifropc, line 244: So, when transaction is in the watchlist, we always call onMessage with this transaction, even if status didn't change, right?
What's the use-case for that?

src/walletSdk/Anchor/index.ts
CassioMG: I'll try moving to a separate class

src/walletSdk/Anchor/index.ts
CassioMG: Kind of, we always call `onMessage` on the **first try** for watchlist transactions, but on subsequent retries we only call `onMessage` again [if there is a change in the transaction](https://github.com/stellar/typescript-wallet-sdk/pull/12/files#diff-08c9260268921699d5ac092cdf53654f59685a9378df00a964422ec206c4c419R318) otherwise it does nothing (i.e. keep silently waiting until there is a transaction change).

Humm I don't have any specific/strong use-case for it so far, but thought it could be useful for wallets that wish to watch for several transactions on an id-basis instead of a status-basis.

That being said, I'm not against removing it in case we'd like to keep it simple for now (and add it later on the road in case we get some clear use cases for it).

Ifropc: Got it. I don't have a strong preference, it's up to you if you want to remove it or keep it.
Agree that keeping it simple for now is a good idea, but leaving it be is also OK with me

src/walletSdk/Anchor/index.ts
CassioMG: @Ifropc just moved it to a `Watcher` class, I'll work on updating the [other PR](https://github.com/stellar/typescript-wallet-sdk/pull/14) now

src/walletSdk/Watcher/index.ts
```diff
@@ -0,0 +1,212 @@
...
+
+export class Watcher {
+  private anchor: Anchor;
+
+  private _allTransactionsWatcher?: ReturnType<typeof setTimeout>;
+  private _watchAllTransactionsRegistry: WatchAllTransactionsRegistry;
+  private _transactionsRegistry: TransactionsRegistry;
+  private _transactionsIgnoredRegistry: TransactionsRegistry;
+
+  constructor(anchor: Anchor) {
```
Ifropc, line 26: Can you add a comment that this class should not be created directly, and instead `.watcher()` on anchor object should be used?

Ifropc (approved):

src/walletSdk/Watcher/index.ts
CassioMG: Done! I used the same comment pattern from Anchor class declaration

PR #13, opened 2023-05-29
add wallet signer class

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-813)

mimicking how the [kotlin-sdk](https://github.com/stellar/kotlin-wallet-sdk/blob/7ef34b4c3d3cae84109c806bc92c9a9fb10cbf68/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/auth/WalletSigner.kt#L10) is handling wallet signing. Creating a generic interface that the user uses to handle signing with client and optional domain account. An example how it can be used is in the test

### DIFF
--- a/src/walletSdk/Auth/index.ts
+++ b/src/walletSdk/Auth/index.ts
@@ -38,7 +38,11 @@ export class Auth {
     return await this.getToken(signedTx);
   }
 
-  async challenge(accountKp: Keypair, memoId?: string, clientDomain?: string) {
+  private async challenge(
+    accountKp: Keypair,
+    memoId?: string,
+    clientDomain?: string
+  ) {
     if (memoId && parseInt(memoId) < 0) {
       throw new InvalidMemoError();
     }
@@ -56,7 +60,7 @@ export class Auth {
     }
   }
 
-  sign(
+  private sign(
     accountKp: Keypair,
     challengeResponse: {
       transaction: string;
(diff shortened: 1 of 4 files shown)


test/index.test.ts
```diff
@@ -56,6 +55,42 @@ describe("Anchor", () => {
...
+          transactionXDR,
+          networkPassPhrase
+        );
+        transaction.sign(clientDomainKp);
+        signedByDomain = true;
+        return transaction;
+      },
+    };
+
+    const challengeResponse = await auth.challenge(
```
Ifropc, line 84: Can you just call auth.authenticate and check that token has been returned?

src/walletSdk/Auth/index.ts
```diff
@@ -52,13 +56,31 @@ export class Auth {
     }
   }
 
-  // TODO - add signing with client account functionality
-  sign(accountKp: Keypair, challengeTx, network) {
-    const transaction = StellarSdk.TransactionBuilder.fromXDR(
-      challengeTx,
-      network
+  sign(
```
Ifropc, line 59: Both this and challenge functions should be private

acharb: 👍 https://github.com/stellar/typescript-wallet-sdk/pull/13/commits/27ec5a5630f9b175df0d6c59d75a24b8eef9cbb4

CassioMG (approved):

test/index.test.ts
acharb: it's because the "demo-wallet-server.stellar.org" domain secret key I don't have (just using a random secret key up at line 71), so it would fail to authenticate. So I'm testing up to that point.

But I don't think we should add it to the test though just to be safe? wdyt?

acharb: added a comment there to make more clear

piyalbasu (approved):

PR #14, opened 2023-05-29
Add watchOneTransaction functionality

CassioMG:
Related task: [[TS] Functionality for watching transactions updates](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-836)


src/walletSdk/Anchor/index.ts
```diff
@@ -401,6 +409,130 @@ export class Anchor {
...
+  * * onError - When there's a runtime error, or the transaction comes back as
+  * no_market / too_small / too_large / error.
+  */
+  watchOneTransaction(params: {
+    authToken: string;
+    assetCode: string;
+    id: string;
+    onMessage: (transaction) => void;
+    onSuccess: (transaction) => void;
+    onError: (error: any) => void;
```
Ifropc, line 425: error: Error | Transaction
?

src/walletSdk/Anchor/index.ts
```diff
@@ -401,6 +409,130 @@ export class Anchor {
...
   }
 
+  /**
+  * Watch a transaction until it stops pending. Takes three callbacks:
+  * * onMessage - When the transaction comes back as pending_ or incomplete.
+  * * onSuccess - When the transaction comes back as completed / refunded / expired.
+  * * onError - When there's a runtime error, or the transaction comes back as
+  * no_market / too_small / too_large / error.
+  */
+  watchOneTransaction(params: {
```
Ifropc, line 419: Let's move it to new watcher class after all. Considering that we now have 2 methods.
It feels like anchor class has way too much things going on right now

src/walletSdk/Anchor/index.ts
CassioMG: Nice catch, the type should be `Error | Transaction`

I didn't put much focus on Typing right now as we have a [dedicated task](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-790) for it where the Transaction type should be created

I think I'll just remove the `any` Type for now and let Typescript infer it

src/walletSdk/Anchor/index.ts
CassioMG: Sounds good to me, I'll try creating a Watcher class

Ifropc (approved):

PR #16, opened 2023-05-31
Add language support

CassioMG:
Related: [[TS] Add lang support for all Sep-24 endpoints](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-837)

This PR adds `language` support to the Wallet (and Anchor) class which should be used as default `lang` param value for all Sep-24 API calls.

Devs can still optionally set the `lang` parameter individually per function.

### DIFF
--- a/src/walletSdk/index.ts
+++ b/src/walletSdk/index.ts
@@ -10,9 +10,9 @@ import { getUrlDomain } from "./util/url";
 // TODO - https://stellarorg.atlassian.net/browse/WAL-789?atlOrigin=eyJpIjoiZjcwNzZhOTJmNjE1NGRhNTk1NDlkOTExMTYxMDJkZmYiLCJwIjoiaiJ9
 interface ClientConfigFn {}
 interface ClientConfig {}
-class HttpClient {}
+export class HttpClient {}
 
-class Config {
+export class Config {
   app: ApplicationConfiguration;
   stellar: StellarConfiguration;
   constructor(stellarCfg, appCfg) {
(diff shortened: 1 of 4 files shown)


src/walletSdk/index.ts
```diff
@@ -35,19 +36,28 @@ export class Wallet {
 
   constructor(
     stellarConfiguration: StellarConfiguration,
-    applicationConfiguration: ApplicationConfiguration = new ApplicationConfiguration()
+    applicationConfiguration: ApplicationConfiguration = new ApplicationConfiguration(),
+    // Defaults wallet language to "en", this will reflect in all Anchor API calls
+    language: string = "en",
```
Ifropc, line 41: Is there a difference between "en" and "en-US"?

CassioMG: @Ifropc so I left the default value as `en` because the [SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#request) mentions it. But, some Anchors (like MGI) may require the wallet to send it using the locale format like `en-US`, because the same language could vary depending on the country. E.g. we could have `en-US`, `en-UK`, `es-AR`, `es-ES`, `pt-BR`, `pt-PT`, etc...

Does that make sense?

CassioMG: If you notice on the tests I always used it as `en-US` just for devs to keep in mind that the `lang` parameter could be used with this `language-locale` format

Ifropc: Yeah I was just curious if "en" and "en-US" are the same

CassioMG: I believe it'll depend more on the Anchor but probably yes :)

PR #17, opened 2023-06-01
Add setAuthToken function

CassioMG:
This PR proposes a way of conveniently persisting the authentication token so that devs don't need to pass it on every function.

If an `authToken` is set on the Anchor object (through `setAuthToken`) then it should be used as default value for `authToken` in all functions.


Ifropc:
I don't really like this change in a sense that logically, anchor is just a connection to the anchor.
Maybe we can add one more class, that will represent logical connection of a particular user to the anchor? (Using token)

PR #19, opened 2023-06-02
Lang param should be optional

CassioMG:
Forgot to make it optional on the getTransactionBy function

### DIFF
--- a/src/walletSdk/Anchor/index.ts
+++ b/src/walletSdk/Anchor/index.ts
@@ -122,7 +122,7 @@ export class Anchor {
     id?: string;
     stellarTransactionId?: string;
     externalTransactionId?: string;
-    lang: string;
+    lang?: string;
   }) {
     if (!id && !stellarTransactionId && !externalTransactionId) {
       throw new MissingTransactionIdError();


Ifropc (approved):

src/walletSdk/Anchor/index.ts
```diff
@@ -122,7 +122,7 @@ export class Anchor {
     id?: string;
     stellarTransactionId?: string;
     externalTransactionId?: string;
-    lang: string;
+    lang?: string;
   }) {
```
Ifropc, line 126: Do we usually specify return type too? Or is it optional?

CassioMG: It's optional but I think it's good practice to specify the return type. I believe once we have [this task](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?modal=detail&selectedIssue=WAL-790) done we could probably revisit all functions and add proper `input` and `output` types.

What do you think @Ifropc @piyalbasu @acharb ?

acharb: I thought it was required by typescript. But even if not I think it helps with readability

Ifropc: I agree, specifying types are much better IMO.
Let's try adding return types to all new functions (and functions that have been changed), and then re-visit it later in the scope of WAL-790
WDYT?

piyalbasu: return types are usually inferred by Typescript, but I think for an SDK, it's good practice to be explicit about return types.

I think it'd be a great idea to explicate return types. I'm going to create a new ticket for this since 790 is already underway

piyalbasu (approved):

PR #20, opened 2023-06-06
add anchor types

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-790?atlOrigin=eyJpIjoiNDliODYwNmY3OWRiNGRkNzkwZjE3OTdmNTBlZTY1YzgiLCJwIjoiaiJ9)

### DIFF
--- a/src/walletSdk/Anchor/Types.ts
+++ b/src/walletSdk/Anchor/Types.ts
@@ -1,4 +1,5 @@
 import { MemoType } from "stellar-sdk";
+import { Optional } from "utility-types";
 import { TransactionStatus } from "../Watcher/Types";
 
 export interface AnchorServiceInfo {
@@ -48,16 +49,8 @@ export interface WithdrawTransaction extends ProcessingAnchorTransaction {
   withdraw_anchor_account: string;
 }
 
-interface ErrorTransaction extends ProcessingAnchorTransaction {
-  from?: string;
-  to?: string;
-  deposit_memo?: string;
-  deposit_memo_type?: MemoType;
-  claimable_balance_id?: string;
-  withdraw_memo?: string;
-  withdraw_memo_type?: MemoType;
-  withdraw_anchor_account?: string;
-}
+interface ErrorTransaction
+  extends Optional<DepositTransaction & WithdrawTransaction> {}
 
 export type AnchorTransaction =
   | DepositTransaction
--- a/src/walletSdk/Anchor/Types.ts
+++ b/src/walletSdk/Anchor/Types.ts
@@ -1,0 +1,86 @@
+import { MemoType } from "stellar-sdk";
+import { TransactionStatus } from "../Watcher/Types";
+
+export interface AnchorServiceInfo {
+  deposit: { [key: string]: AnchorServiceAsset };
+  withdraw: { [key: string]: AnchorServiceAsset };
+  fee: { enabled: boolean };
+  features: { account_creation: boolean; claimable_balances: boolean };
+}
+
+interface BaseTransaction {
+  id: string;
+  status: TransactionStatus;
+  more_info_url: string;
+  started_at: string;
+  message?: string;
+}
+
+interface ProcessingAnchorTransaction extends BaseTransaction {
+  status_eta?: number;
+  kyc_verified?: boolean;
+  amount_in_asset?: string;
+  amount_in: string;
+  amount_out_asset?: string;
+  amount_out: string;
+  amount_fee_asset?: string;
+  amount_fee: string;
+  completed_at?: string;
+  stellar_transaction_id?: string;
+  external_transaction_id?: string;
+  refunds?: Refunds;
+}
+
+export interface DepositTransaction extends ProcessingAnchorTransaction {
+  from?: string;
+  to?: string;
+  deposit_memo?: string;
+  deposit_memo_type?: MemoType;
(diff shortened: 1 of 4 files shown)


src/walletSdk/Anchor/Types.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface AnchorServiceInfo {
+  deposit: { [key: string]: AnchorServiceAsset };
+  withdraw: { [key: string]: AnchorServiceAsset };
+  fee: { enabled: boolean };
+  features: { account_creation: boolean; claimable_balances: boolean };
+}
+
+interface BaseTransaction {
+  id: string;
```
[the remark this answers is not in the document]
acharb: yup good catch https://github.com/stellar/typescript-wallet-sdk/pull/20/commits/5abb4d4caad8e89ccd51954a961bc1fb3664b17a

@Ifropc it looks it's also missing from the kotlin sdk (if I'm reading it right)

src/walletSdk/Anchor/Types.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface WithdrawTransaction extends ProcessingAnchorTransaction {
+  from: string;
+  to?: string;
+  withdraw_memo?: string;
+  withdraw_memo_type: MemoType;
+  withdraw_anchor_account: string;
+}
+
+interface ErrorTransaction extends ProcessingAnchorTransaction {
+  kind: string;
```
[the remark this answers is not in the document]
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/20/commits/5abb4d4caad8e89ccd51954a961bc1fb3664b17a

src/walletSdk/Anchor/Types.ts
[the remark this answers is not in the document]
Ifropc: Yep kind would be nice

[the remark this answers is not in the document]
Ifropc: Oh I think I remembered, kind is only in error transaction, because all other transactions are already typed (it's either Deposit/Withdrawal/IncompleteDeposit/IncompleteWithdrawal)
The reason for incomplete is because they are missing a lot of fields, that are mandatory for any in-progress status transaction.
Not sure if it's user-friendly for typescript. 
In Kotlin, I can write
```kotlin
val tx: AnchorTransaction
when (tx) -> 
 is Deposit -> // handle deposit
 is IncompleteDeposit -> // handle incompelte deposit
 is Error -> // handle error
 else -> Exception("Unexpected kind") 
```

src/walletSdk/Anchor/Types.ts
```diff
@@ -0,0 +1,86 @@
...
+  amount_fee: string;
+  completed_at?: string;
+  stellar_transaction_id?: string;
+  external_transaction_id?: string;
+  refunds?: Refunds;
+}
+
+export interface DepositTransaction extends ProcessingAnchorTransaction {
+  from?: string;
+  to?: string;
```
CassioMG, line 37: Based on the [SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#fields-for-deposit-transactions) it looks like `from` and `to` are required

src/walletSdk/Anchor/Types.ts
```diff
@@ -0,0 +1,86 @@
...
+  deposit_memo_type?: MemoType;
+  claimable_balance_id?: string;
+}
+
+export interface WithdrawTransaction extends ProcessingAnchorTransaction {
+  from: string;
+  to?: string;
+  withdraw_memo?: string;
+  withdraw_memo_type: MemoType;
+  withdraw_anchor_account: string;
```
CassioMG, line 48: All params on `WithdrawTransaction` should be required according to the [SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#fields-for-withdraw-transactions)

src/walletSdk/Anchor/Types.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface ErrorTransaction extends ProcessingAnchorTransaction {
```
CassioMG, line 51: It looks like this `ErrorTransaction`  interface is a merge of `DepositTransaction & WithdrawTransaction` interfaces

I believe we could do something like this on TypeScript which would result in the same thing:
`export interface ErrorTransaction extends DepositTransaction, WithdrawTransaction {}`

:)

src/walletSdk/Anchor/Types.ts
Ifropc: Workaround for MGI...

src/walletSdk/Anchor/Types.ts
Ifropc: Same here, memo is missing sometimes for MGI

src/walletSdk/Anchor/Types.ts
Ifropc: Good idea!

acharb: 👍 https://github.com/stellar/typescript-wallet-sdk/pull/20/commits/8078bdc5d2308dcc705fdac931a9b49ca9cca380

CassioMG (approved):

PR #21, opened 2023-06-13
add keypair code

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-761?atlOrigin=eyJpIjoiYjU2YTliYzgyNjdmNDhiM2IxYWUxMzZjNjZlZTVlMGEiLCJwIjoiaiJ9)

left some comments on design decisions, happy to discuss if needd


src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  static fromSecret = (secret: string): SigningKeypair => {
+    return new SigningKeypair(Keypair.fromSecret(secret));
+  };
+
+  get secretKey(): string {
+    return this.keypair.secret();
+  }
+
+  sign(
```
acharb, line 42: I played around with extending the stellar-sdk Transaction class using augmentation in Typescript (to make similar to [kotlin here](https://github.com/stellar/kotlin-wallet-sdk/blob/e6c801e24e7c716b876a67098131dd0b4d2f1691/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Account.kt#L46)). It gets pretty messy imo. So I left just the `sign` method here for signing with keypair

to sign with the transaction someone could just do: `tx.sign(signingKeypair.keypair)`

src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,52 @@
...
+    transaction: Transaction | FeeBumpTransaction
+  ): Transaction | FeeBumpTransaction {
+    transaction.sign(this.keypair);
+    return transaction;
+  }
+
+  toString(): string {
+    return `${this.constructor.name}(address=${this.publicKey})`;
+  }
+}
```
acharb, line 52: similarly I didn't try to extend the String class like kotlin does [here](https://github.com/stellar/kotlin-wallet-sdk/blob/e6c801e24e7c716b876a67098131dd0b4d2f1691/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Account.kt#L50)

imo extending native classes may lead to some wonky behavior, but happy to discuss

src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,52 @@
+import { Keypair, Transaction, FeeBumpTransaction } from "stellar-sdk";
+
+class AccountKeypair {
+  keypair: Keypair;
+  constructor(keypair: Keypair) {
+    this.keypair = keypair;
+  }
+
+  get publicKey(): string {
```
acharb, line 9: @Ifropc  is there a reason to include both `publicKey` and `accountId` from the keypair? https://github.com/stellar/kotlin-wallet-sdk/blob/e6c801e24e7c716b876a67098131dd0b4d2f1691/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Account.kt#L13-L16

or can we use just `publicKey`, I'm unsure of the use case for having `accountId` as well 🤔

Ifropc: `accountId` is very useful because it's human-readable
in fact, in most places we use `accountId`, and `publicKey` is not used that much

src/walletSdk/horizon/Account.ts
Ifropc: Yeah I'm guessing all extensions functions we have on kotlin side should either be a part of original class, or if it extends some external functionality you can just use util functions

src/walletSdk/horizon/Account.ts
Ifropc: Yeah adding method `PublicKeyPair.fromString()` should be good enough

src/walletSdk/horizon/AccountService.ts
```diff
@@ -0,0 +1,18 @@
...
+  private server;
+  private network;
+
+  constructor(cfg) {
+    this.server = cfg.stellar.server;
+    this.network = cfg.stellar.network;
+  }
+
+  createKeypair() {
+    return new SigningKeypair(Keypair.random());
```
Ifropc, line 16: This won't work on react native. Not sure what's the best way of working it around, maybe adding typescript-wallet-sdk-react-native module later?
Reference: https://github.com/stellar/js-stellar-sdk#using-in-an-expo-managed-workflow

piyalbasu: Interesting - it looks like using `expo-random` to generate a keypair should work in the browser. Should we just default to that? It'd be great if we didn't have to branch logic for react native (not sure if there are other areas where we definitely have to do that)

piyalbasu: ah yeah, it looks like it doesn't run in node. Yeah, we could do a try with the native `crypto` lib for node and fallback to `expo-crypto` (that's actually the new lib, expo-random has been deprecated)

Ifropc: IMO -- let's do it later. We should add some basic e2e tests for all platforms we want to support (browser, node, RN)

src/walletSdk/horizon/Account.ts
acharb: woops I mixed them up, I meant unsure of the `publicKey` use case. Do we need both? (ie do we need the ByteArray one?)

src/walletSdk/horizon/Account.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/2f244a59a2922afac45f7e3cb9aaeded5d062a72

src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,57 @@
...
+    super(keypair);
+  }
+
+  static fromString = (str: string) => {
+    const kp = Keypair.fromPublicKey(str);
+    return new PublicKeypair(kp);
+  };
+}
+
+export class SigningKeypair extends AccountKeypair {
```
CassioMG, line 30: should it extend `PublicKeypair` instead? I'd assume having a secret gives you the power to have all `PublicKeypair` functions as well

src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+export class PublicKeypair extends AccountKeypair {
+  keypair: Keypair;
+  constructor(keypair: Keypair) {
+    super(keypair);
+  }
+
+  static fromString = (str: string) => {
```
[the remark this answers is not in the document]
CassioMG: maybe also use the same code pattern you used on the `fromSecret` function:
```
  static fromPublicKey = (key: string): PublicKeypair => {
    return new PublicKeypair(Keypair.fromPublicKey(key));
  };
```

:)

src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,57 @@
...
+  get publicKey(): string {
+    return this.keypair.publicKey();
+  }
+
+  toString(): string {
+    return `${this.constructor.name}(address=${this.publicKey})`;
+  }
+}
+
+export class PublicKeypair extends AccountKeypair {
```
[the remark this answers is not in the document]
CassioMG: given that we have for `AccountKeypair` and `SigningKeypair`

src/walletSdk/horizon/AccountService.ts
```diff
@@ -0,0 +1,18 @@
+import { Keypair } from "stellar-sdk";
+
+import { SigningKeypair } from "./Account";
+
+// Do not create this object directly, use the Wallet class.
+export class AccountService {
+  private server;
+  private network;
```
CassioMG, line 8: could we add types for `server` and `network` ?

src/walletSdk/horizon/Stellar.ts
```diff
@@ -1,6 +1,13 @@
-// TODO - https://stellarorg.atlassian.net/browse/WAL-814?atlOrigin=eyJpIjoiMWY5MjBkZTE4MTg3NDA3N2E0MjYwMmQ2ZmRhOGFiODUiLCJwIjoiaiJ9
+import { AccountService } from "./AccountService";
 
 // Do not create this object directly, use the Wallet class.
 export class Stellar {
-  constructor(cfg) {}
+  private cfg;
```
CassioMG, line 5: could we add type for `cfg` ?

src/walletSdk/horizon/AccountService.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  constructor(cfg) {
+    this.server = cfg.stellar.server;
+    this.network = cfg.stellar.network;
+  }
+
+  createKeypair() {
```
CassioMG, line 15: could we add return type here?

src/walletSdk/horizon/Account.ts
Ifropc: In Kotlin, we have them split. So all methods either accept `AccountKeyPair` (so either will work), or strictly `SigningKeyPair`. 
I think your proposal is more or less the same as what we have right now, but probably better to keep differences as small as possible.

src/walletSdk/horizon/Account.ts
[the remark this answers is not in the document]
Ifropc: Some people also call it account or address, could be confusing. In kotlin we simply have `"myString".toPublicKeyPair()` :)
Here, don't know what's the right naming is, but I like `PublicKeyPair.fromString()`

src/walletSdk/horizon/Account.ts
Ifropc: I don't think we need the ByteArray. It was originally added there to match the SDK KeyPair, but I think we can omit it

src/walletSdk/horizon/Account.ts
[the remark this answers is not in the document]
acharb: ooh good point, I vote for `fromPublicKey` actually to be in contrast of `fromSecret` from SigningKeypair

https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/30f981ef80dd2bbd93cd931ae9df459f3d1afc66

src/walletSdk/horizon/Account.ts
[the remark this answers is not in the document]
acharb: ah good catch. We don't need the overwrites actually, removed SigningKeypair's

https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/de4fab96cc3907225016cfe295e48127890ab6e5

src/walletSdk/horizon/AccountService.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/387e2f988d5d91f79734e0f27eceaf8662427fb2

src/walletSdk/horizon/Stellar.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/44202897d86f2a6b9f27384178bfdae68b9d4646

src/walletSdk/horizon/AccountService.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/a281475ea5eb914e55a7d616a969882d87aeda3e

src/walletSdk/horizon/AccountService.ts
acharb: I think this'll take some more digging into. I created a ticket to track here: https://stellarorg.atlassian.net/browse/WAL-885

going to leave for now if that's alright

src/walletSdk/horizon/Account.ts
```diff
@@ -0,0 +1,52 @@
...
+  keypair: Keypair;
+  constructor(keypair: Keypair) {
+    super(keypair);
+  }
+
+  static fromPublicKey = (str: string) => {
```
CassioMG, line 24: could we add return type here as well?

sorry for being annoying about the typing, I'm just trying to keep consistency throughout the codebase :sweat_smile:

maybe we should have a task dedicated to reviewing and adding types everywhere

CassioMG (approved):
Lgtm, left a comment regarding typing

Ifropc (approved):

src/walletSdk/horizon/Account.ts
acharb: not a problem glad we're catching / adding these 😄
https://github.com/stellar/typescript-wallet-sdk/pull/21/commits/f4dbf01c448df491da717687e4591e393b835d29

PR #24, opened 2023-07-11
use AccountKeypair in sdk

acharb:
when adding docs in https://github.com/stellar/typescript-wallet-sdk/pull/25 noticed 3 changes need to be made:
1. use `AccountKeypair` instead of the stellar-sdk Keypair class
2. WalletSigner.signWithDomainAccount needs to be async since meant to contain a backend call
3. be able to pass in a custom defined client

### DIFF
--- a/src/walletSdk/index.ts
+++ b/src/walletSdk/index.ts
@@ -59,7 +59,7 @@ export class Wallet {
     return new Anchor({
       cfg: this.cfg,
       homeDomain: getUrlDomain(url),
-      httpClient: this.getClient(httpClientConfig),
+      httpClient: this.cfg.app.defaultClient,
       language,
     });
   }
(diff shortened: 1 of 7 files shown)


src/walletSdk/index.ts
```diff
@@ -75,20 +75,10 @@ export class Wallet {
     return new Recovery({
       cfg: this.cfg,
       stellar: this.stellar(),
-      httpClient: this.getClient(httpClientConfig),
+      httpClient: this.cfg.app.defaultClient,
       servers
     });
   }
-
-  getClient(httpClientConfig: AxiosRequestConfig = {}): AxiosInstance {
```
acharb, line 83: originally this method was added because kotlin had it. After looking at it again it's because the kotlin sdk needs to create and store multiple client instances so it can close them later, which typescript sdk doesn't have to do. So removing this method

acharb: @Ifropc lmk if my understanding of why the kotlin sdk has this is wrong

Ifropc (approved):

PR #27, opened 2023-07-18
reorganize sep-24 code

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-945?atlOrigin=eyJpIjoiZWI4ODdkYjIyM2E4NDk5NmI2Y2ZmZWIwYmI0ODJkYTAiLCJwIjoiaiJ9)

making the code structure changes that were made in the [kotlin sdk](https://github.com/stellar/kotlin-wallet-sdk/pull/107)


src/walletSdk/Watcher/index.ts
```diff
@@ -112,79 +112,78 @@ export class Watcher {
...
     }
 
-    this.anchor.getTransactionsForAsset({ 
-      authToken, 
-      assetCode, 
-      lang, 
-      kind, 
-      noOlderThan 
-    }).then((transactions: AnchorTransaction[]) => {
+    this.anchor
```
Ifropc, line 115: Instead of Anchor you can pass Sep24 class in the constructor now

docs/WalletGuide.md
```diff
@@ -64,7 +64,7 @@ let anchor = wallet.anchor({ homeDomain: "testanchor.stellar.org" });
 Get anchor information from a TOML file.
```
Ifropc, line 64: I think we should remove this md file in favor of stellar-docs (maybe part of WAL-972)

Ifropc (approved):
LGTM

src/walletSdk/Watcher/index.ts
acharb: thoughts on keeping it as anchor? Since the watcher needs to access the language field (and wouldn't be able to sep24.anchor.language since anchor is private)

Ifropc: Good point

PR #28, opened 2023-07-18
add prettier and husky pre-commit

acharb:
adding js prettier so we have a standard on code formatting/style


Shaptic (commented):
You can use lint-staged (recommended by prettier: https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged) to hook into linting alongside husky, though I don’t really see how it’s an advantage over just husky+bash as you’ve done. Maybe the UI is prettier?

piyalbasu:
> You can use lint-staged (recommended by prettier: https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged) to hook into linting alongside husky, though I don’t really see how it’s an advantage over just husky+bash as you’ve done. Maybe the UI is prettier?

Interestingly, lint-staged uses Husky under the hood, so you'd end up with something similar to what Alec has done anyway. lint-staged just gives you a bunch of config options (running multiple jobs pre-commit in parallel, for ex) that we don't really need. So, I think it's fine to just set up the Husky cmd ourselves so we have a little more control over it until we need something more robust.

Also fwiw, most of us use a prettier config in our IDE's that auto formats on save anyway

piyalbasu (approved):

piyalbasu:
lgtm once the conflicts are resolved 👍

PR #29, opened 2023-07-24
improve auth token handling

acharb:


src/walletSdk/Types/sep24.ts
```diff
@@ -13,11 +13,12 @@ export type ExtraFields = {
...
 export type Sep24PostParams = {
   authToken: AuthToken;
-  accountAddress: string;
   assetCode: string;
   lang?: string;
   extraFields?: ExtraFields;
-  fundsAccountAddress?: string;
+  destinationAccount?: string;
+  withdrawalAccount?: string;
+  account?: string;
```
CassioMG, line 21: could we remove `destinationAccount + withdrawalAccount` altogether and just leave `account` for simplicity?

CassioMG (approved):
LGTM - left a suggestion regarding redundancy on `Sep24PostParams` params

src/walletSdk/Types/sep24.ts
acharb: ya we could, I thought to leave them though to keep the interface consistent with the [kotlin sdk](https://github.com/stellar/kotlin-wallet-sdk/blob/77e4bfe0eae19b2ede5af231147351f8367cef01/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Sep24.kt#L45)

I think it does make clearer to the user as well what the account is doing

PR #30, opened 2023-07-26
allow deposit memo in deposits

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-912?atlOrigin=eyJpIjoiMzYwZGEwYjI1MDk3NDIwOGE3YmNkNTg4OTEyNjZiOTYiLCJwIjoiaiJ9)

realized missing when updating stellar docs in https://github.com/stellar/stellar-docs/pull/196

### DIFF
--- a/src/walletSdk/Anchor/Sep24.ts
+++ b/src/walletSdk/Anchor/Sep24.ts
@@ -46,6 +46,7 @@ export class Sep24 {
     authToken,
     lang,
     extraFields,
+    destinationMemo,
     destinationAccount,
   }: Sep24PostParams): Promise<Sep24PostResponse> {
     return this.flow({
@@ -53,6 +54,7 @@ export class Sep24 {
       authToken,
       lang,
       extraFields,
+      destinationMemo,
       account: destinationAccount,
       type: FLOW_TYPE.DEPOSIT,
     });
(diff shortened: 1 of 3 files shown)


src/walletSdk/Anchor/Sep24.ts
```diff
@@ -46,13 +46,15 @@ export class Sep24 {
...
     extraFields,
+    destinationMemo,
     destinationAccount,
   }: Sep24PostParams): Promise<Sep24PostResponse> {
     return this.flow({
       assetCode,
       authToken,
       lang,
       extraFields,
+      destinationMemo,
```
CassioMG, line 57: Nice addition. Should we also add memo to `withdraw()` ? According to the SEP it should be possible: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#withdraw

acharb: Ya i noticed that too but the Kotlin sdk didn't have it so I left it off

@Ifropc any reason for that? or can I go ahead and memos for withdrawals too

Ifropc: It's deprecated and should be derived from SEP-10 token instead. I don't know of any anchor that doesn't support this functionality

Ifropc (approved):

src/walletSdk/Anchor/Sep24.ts
CassioMG: Makes sense, I've only seen cases for deposit memo

CassioMG (approved):

PR #31, opened 2023-07-31
add submitWithFeeIncrease

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-774?atlOrigin=eyJpIjoiYTJhNzY2NjYyODRhNDExNTliYzFjNjQ4ZjNhNjRlOTQiLCJwIjoiaiJ9)

### DIFF
--- a/src/walletSdk/Horizon/Stellar.ts
+++ b/src/walletSdk/Horizon/Stellar.ts
@@ -1,12 +1,20 @@
-import { Account as StellarAccount, Server, Transaction } from "stellar-sdk";
+import {
+  Account as StellarAccount,
+  Server,
+  Transaction,
+  Memo,
+} from "stellar-sdk";
+
 import { Config } from "walletSdk";
 import { AccountService } from "./AccountService";
 import { TransactionBuilder } from "./transaction/TransactionBuilder";
 import { TransactionParams } from "../Types";
+import { AccountKeypair } from "./Account";
 import {
   AccountDoesNotExistError,
   TransactionSubmitFailedError,
 } from "../Exceptions";
+import { getResultCode } from "../Utils/getResultCode";
 
 // Do not create this object directly, use the Wallet class.
 export class Stellar {
@@ -69,6 +77,23 @@ export class Stellar {
         // https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/timeout
         // https://developers.stellar.org/docs/encyclopedia/error-handling#timeouts
         return await this.submitTransaction(signedTransaction);
+      }
+      throw e;
+    }
+  }
+
+  async submitWithFeeIncrease(
+    signedTransaction: Transaction,
+    baseFeeIncrease: number,
+  ) {
+    try {
+      const resp = await this.submitTransaction(signedTransaction);
+      return signedTransaction;
+    } catch (e) {
+      const resultCode = getResultCode(e);
+      if (resultCode === "tx_too_late") {
+        const newFee = parseInt(signedTransaction.fee) + baseFeeIncrease;
+        return this.submitWithFeeIncrease(signedTransaction, baseFeeIncrease);
       }
       throw e;
     }
--- a/src/walletSdk/Horizon/Stellar.ts
+++ b/src/walletSdk/Horizon/Stellar.ts
@@ -13,8 +13,10 @@ import {
   AccountDoesNotExistError,
   TransactionSubmitFailedError,
   TransactionSubmitWithFeeIncreaseFailedError,
+  SignerRequiredError,
 } from "../Exceptions";
 import { getResultCode } from "../Utils/getResultCode";
+import { SigningKeypair } from "./Account";
 
 // Do not create this object directly, use the Wallet class.
 export class Stellar {
@@ -87,7 +89,7 @@ export class Stellar {
     timeout,
     baseFeeIncrease,
     operations,
-    signingAddresses,
+    signerFunction,
     baseFee,
     memo,
     maxFee,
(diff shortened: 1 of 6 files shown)


src/walletSdk/Horizon/Stellar.ts
```diff
@@ -73,4 +81,21 @@ export class Stellar {
       throw e;
     }
   }
+
+  async submitWithFeeIncrease(
```
acharb, line 85: @Ifropc I simplified a bit from what [kotlin is doing](https://github.com/stellar/kotlin-wallet-sdk/blob/77e4bfe0eae19b2ede5af231147351f8367cef01/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Stellar.kt#L161-L230) here. For two reasons:

1. keeps it consistent with how the regular `submitTransaction` works
2. there's not a great way in typescript to do the lambda function thing that kotlin is doing [here](https://github.com/stellar/kotlin-wallet-sdk/blob/77e4bfe0eae19b2ede5af231147351f8367cef01/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Stellar.kt#L203) for `buildingFunction`. In Typescript we'd have to create a new TransactionBuild object w/in the function, which would be a bit extra complexity we wouldn't need imo (Unless there's a reason `submitWithFeeIncrease` needs to create a new transaction)

wdyt?

Ifropc: Current implementation doesn't match Kotlin one.
There is reason it needs to create a new transaction. It bumps fee on every step. 
So, first transaction is submitted with a fee of 100
Next transaction is submitted with a fee of 11000 and so on (assuming fee increase is 1000)
Without this mechanism transaction submitting is not reliable during network surges

Ifropc: BTW, we should also add maxFee to this method

acharb: yup I see that now, totally whiffed on that

fixed with adding maxFee: https://github.com/stellar/typescript-wallet-sdk/pull/31/commits/bbd0f8753535d4ccd5208a801943ced3ad7df5c0

src/walletSdk/Horizon/Stellar.ts
```diff
@@ -73,4 +81,60 @@ export class Stellar {
...
+      baseFee,
+      memo,
+    });
+
+    operations.forEach((op) => {
+      builder.addOperation(op);
+    });
+
+    const transaction = builder.build();
+    if (signingAddresses?.length) {
```
acharb, line 107: @Ifropc what's the use case for the [signer function in kotlin here](https://github.com/stellar/kotlin-wallet-sdk/blob/77e4bfe0eae19b2ede5af231147351f8367cef01/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Stellar.kt#L202)? I'm imagining it's for signing with one or multiple other keypairs, but lmk if that's wrong

Ifropc: It's to sign the transaction. In general, it could be not trivial, so I left it for the developer to implement

src/walletSdk/Horizon/Stellar.ts
```diff
@@ -73,4 +81,60 @@ export class Stellar {
...
+      signingAddresses.forEach((signer) => {
+        transaction.sign(signer.keypair);
```
Ifropc, line 109: This won't work for remote signing. Can you add a signer function instead of signing addresses instead please?

src/walletSdk/Types/horizon.ts
```diff
@@ -12,3 +12,14 @@ export type TransactionParams = {
   memo?: Memo;
   timebounds?: Server.Timebounds | number;
 };
+
+export type SubmitWithFeeIncreaseParams = {
+  sourceAddress: SigningKeypair;
```
Ifropc, line 17: Why is it SigningKeyPair? It should be just a regular AccountKeyPair.
We can create a transaction without owning a signing key, send it to the server to sign, and submit. 
It makes sense to do it on the device if the transaction manages device's account, but it's owned by custodial service (for example)

src/walletSdk/Horizon/Stellar.ts
```diff
@@ -73,4 +81,60 @@ export class Stellar {
...
+      });
+    } else {
+      transaction.sign(sourceAddress.keypair);
```
Ifropc, line 112: If function is null AND sourceAddress is SigningKeyPair, we can default to 
`transaction.sign(sourceAddress.keypair);`
otherwise -> throw exception requiring to specify signing function

Ifropc (changes requested):

src/walletSdk/Horizon/Stellar.ts
acharb: 👍  https://github.com/stellar/typescript-wallet-sdk/pull/31/commits/aa70c042ae7b62225fac9122e83a9000c159aac5

src/walletSdk/Types/horizon.ts
acharb: 👍  https://github.com/stellar/typescript-wallet-sdk/pull/31/commits/aa70c042ae7b62225fac9122e83a9000c159aac5

src/walletSdk/Horizon/Stellar.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/31/commits/aa70c042ae7b62225fac9122e83a9000c159aac5

Ifropc (approved):

PR #34, opened 2023-08-07
Manage trustlines

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-769)


src/walletSdk/Asset/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export class NativeAssetId extends StellarAssetId {
+  id = "native";
+  code = "XLM";
+}
+
+export class FiatAssetId extends AssetId {
+  id: string;
+  scheme = FIAT_SCHEME;
+  constructor(id: string) {
```
CassioMG, line 50: What do you think about renaming this constructor input to `code` instead of `id`?
```
  constructor(code: string) {
    super();
    this.id = code;
  }
```

I personally think it'd be more intuitive since for fiat we use code as the id, and asset id always makes me think it's something more complex with a `:` in it

CassioMG (approved):
Lgtm, left a suggest about naming

src/walletSdk/Asset/index.ts
acharb: ya I agree with that, changed https://github.com/stellar/typescript-wallet-sdk/pull/34/commits/434ec91051ce906e07d0e41802609cab8a0b2740

piyalbasu (approved):

Ifropc (commented):
LGTM

PR #36, opened 2023-08-09
add helper method for creating a keypair from random bytes

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-885?atlOrigin=eyJpIjoiYzY2ZDljNzExZmIyNGJjMGFiZGNmMzlkMWFiOWM0ZjMiLCJwIjoiaiJ9)

[original discussion](https://github.com/stellar/typescript-wallet-sdk/pull/21#discussion_r1228424684)

I tried/failed different ways to use [expo-crypto](https://docs.expo.dev/versions/latest/sdk/crypto/#cryptogetrandombytesbytecount) here, to use in a react-native environment. But I wasn't able to get it to install. It requires other react-native environment dependencies (expo-modoule-core, react-native). 

There may be hacky ways to get it to install, but imo it seems messy to install the react-native dependencies that would be installed for all environments (unless there's a clean way of conditionally installing for only react-native that I'm not aware of 🤔)

So I propose this helper method, and then in the documentation we can explain how to use if in a react-native environment:

```typescript
import * as Random from "expo-crypto";
const rand = Random.randomBytes(32);
const kp = account.createKeypairFromRandom(Buffer.from(rand));
```


Ifropc (approved):

CassioMG (approved):

piyalbasu:
@acharb I like this approach as a stopgap. No need to overcomplicate at this stage

@CassioMG @Ifropc I created this spike ticket to investigate how we might want to handle this (and other react-native quirks in the future). I don't think it's high priority, but lmk if anyone disagrees!
https://stellarorg.atlassian.net/browse/WAL-1013?atlOrigin=eyJpIjoiZmQ2NGYxNjkxZDU4NDg2NWI0MzkxYjNjZTMwMWY2ZWUiLCJwIjoiaiJ9

piyalbasu (approved):

PR #38, opened 2023-08-09
Export new classes to wallets

CassioMG:
Exports `IssuedAssetId, NativeAssetId, FiatAssetId and TransactionBuilder` classes.

Every time a new class is created we should add it to `src/index`'s export list so we make sure wallets(clients) will be able to access it.


Ifropc (approved):

piyalbasu (approved):

acharb (approved):
ah good point, will start doing thx for adding 😄

PR #39, opened 2023-08-15
add transfer withdrawal method

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-873?atlOrigin=eyJpIjoiZWFlNGVhNmZkZWMwNGUwNjlkMWMxYTFhZjE3NjNhZmIiLCJwIjoiaiJ9)
The submitWithFeeIncrease method had already been added, but the transfer code hadn't so adding that here

updating the docs in this PR

### DIFF
--- a/src/walletSdk/Horizon/transaction/TransactionBuilder.ts
+++ b/src/walletSdk/Horizon/transaction/TransactionBuilder.ts
@@ -10,8 +10,13 @@ import StellarSdk, {
 
 import { Config } from "walletSdk";
 import { AccountKeypair } from "../Account";
-import { InsufficientStartingBalanceError } from "../../Exceptions";
-import { IssuedAssetId } from "../../Asset";
+import {
+  InsufficientStartingBalanceError,
+  WithdrawalTxMissingMemoError,
+  WithdrawalTxNotPendingUserTransferStartError,
+} from "../../Exceptions";
+import { IssuedAssetId, StellarAssetId } from "../../Asset";
+import { WithdrawTransaction, TransactionStatus } from "../../Types";
 
 export class TransactionBuilder {
   private network: Networks;
@@ -42,7 +47,10 @@ export class TransactionBuilder {
     this.sourceAccount = sourceAccount.accountId();
   }
 
-  createAccount(newAccount: AccountKeypair, startingBalance: number = 1) {
+  createAccount(
+    newAccount: AccountKeypair,
+    startingBalance: number = 1,
+  ): TransactionBuilder {
     if (startingBalance < 1) {
       throw new InsufficientStartingBalanceError();
     }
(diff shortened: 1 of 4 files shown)


src/walletSdk/Horizon/transaction/TransactionBuilder.ts
```diff
@@ -82,4 +114,27 @@ export class TransactionBuilder {
     });
     return this.builder.build();
   }
+
+  transferWithdrawalTransaction(
+    transaction: WithdrawTransaction,
+    assetId: StellarAssetId,
```
CassioMG, line 120: @acharb could we infer the asset from the transaction? I think having an asset should probably be mandatory for a transaction in the `pending_user_transfer_start` status. What do you think?

src/walletSdk/Horizon/transaction/TransactionBuilder.ts
```diff
@@ -82,4 +114,27 @@ export class TransactionBuilder {
...
+  ): TransactionBuilder {
+    if (transaction.status !== TransactionStatus.pending_user_transfer_start) {
+      throw new WithdrawalTxNotPendingUserTransferStartError(
+        transaction.status,
+      );
+    }
+
+    if (!(transaction.withdraw_memo_type && transaction.withdraw_memo)) {
+      throw new WithdrawalTxMissingMemoError();
```
CassioMG, line 129: should we add errors for missing asset and amount fields as well?

CassioMG: unless the `transfer()` function already throw errors for those missing fields 🤔

src/walletSdk/Horizon/transaction/TransactionBuilder.ts
acharb: hmm that's a good question. I'm leaning towards leaving it because according to [sep docs](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#shared-fields-for-both-deposits-and-withdrawals) the amount_in_asset field is optional, so to be safe

the kotlin sdk also has the `assetId` param like this, @Ifropc wdyt?

src/walletSdk/Horizon/transaction/TransactionBuilder.ts
acharb: ya if those are empty creating the payment operation in `transfer` will throw an error, so I don't think we need to add the checks

Ifropc (approved):

src/walletSdk/Horizon/transaction/TransactionBuilder.ts
Ifropc: > could we infer the asset from the transaction? I think having an asset should probably be mandatory for a transaction in the pending_user_transfer_start status

MGI and some other anchors... 
I'm not opposed to have 2 methods tho, one is simplified with just transaction, and the other one is this one

acharb: I'm going to go in favor of simplicity. Imo a second method like `transferWithdrawalTransactionInfer` (since you can't have two methods w/ same name in ts) would be more confusing for the user

if either feel strongly though lmk

PR #45, opened 2023-08-29
add sponsoring

acharb:
adding the ability to sponsor transasctions [ticket](https://stellarorg.atlassian.net/browse/WAL-764?atlOrigin=eyJpIjoiMDBkZGExMGQ0MTgyNGMwODhkYTUxOGU2YjM5ODJhNmEiLCJwIjoiaiJ9)

### DIFF
--- a/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
+++ b/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
@@ -37,19 +37,6 @@ export class SponsoringBuilder extends CommonTransactionBuilder<SponsoringBuilde
     return this;
   }
 
-  addAssetSupport(
-    asset: IssuedAssetId,
-    trustLimit?: string,
-  ): SponsoringBuilder {
-    super.addAssetSupport(asset, trustLimit);
-    return this;
-  }
-
-  removeAssetSupport(asset: IssuedAssetId): SponsoringBuilder {
-    super.removeAssetSupport(asset);
-    return this;
-  }
-
   startSponsoring() {
     this.operations.push(
       StellarSdk.Operation.beginSponsoringFutureReserves({
--- a/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
+++ b/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
@@ -1,0 +1,74 @@
+import StellarSdk, {
+  TransactionBuilder as StellarTransactionBuilder,
+  Transaction,
+} from "stellar-sdk";
+import { IssuedAssetId } from "../../Asset";
+
+import { CommonTransactionBuilder } from "./CommonTransactionBuilder";
+import { AccountKeypair } from "../Account";
+
+export class SponsoringBuilder extends CommonTransactionBuilder<SponsoringBuilder> {
+  private sponsorAccount: AccountKeypair;
+  private builder: StellarTransactionBuilder;
+
+  constructor(
+    sponsoredAddress: string,
+    sponsorAccount: AccountKeypair,
+    builder: StellarTransactionBuilder,
+  ) {
+    super(sponsoredAddress);
+    this.sponsorAccount = sponsorAccount;
+    this.builder = builder;
+
+    this.startSponsoring();
+  }
+
+  createAccount(
+    newAccount: AccountKeypair,
+    startingBalance: number = 0,
+  ): SponsoringBuilder {
+    this.operations.push(
+      StellarSdk.Operation.createAccount({
+        destination: newAccount.publicKey,
+        startingBalance: startingBalance.toString(),
+        source: this.sponsorAccount.publicKey,
+      }),
+    );
+    return this;
+  }
+
+  addAssetSupport(asset: IssuedAssetId, trustLimit?: string) {
+    super.addAssetSupport(asset, trustLimit);
+    return this;
+  }
+
+  removeAssetSupport(asset: IssuedAssetId) {
+    super.removeAssetSupport(asset);
+    return this;
+  }
+
+  startSponsoring() {
+    this.operations.push(
+      StellarSdk.Operation.beginSponsoringFutureReserves({
+        sponsoredId: this.sourceAddress,
+        source: this.sponsorAccount.publicKey,
+      }),
+    );
+  }
+
+  stopSponsoring() {
+    this.operations.push(
+      StellarSdk.Operation.endSponsoringFutureReserves({
+        source: this.sourceAddress,
+      }),
+    );
+  }
+
+  build(): Transaction {
+    this.stopSponsoring();
+    this.operations.forEach((op) => {
+      this.builder.addOperation(op);
+    });
+    return this.builder.build();
+  }
+}
--- a/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
+++ b/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
@@ -37,12 +37,15 @@ export class SponsoringBuilder extends CommonTransactionBuilder<SponsoringBuilde
     return this;
   }
 
-  addAssetSupport(asset: IssuedAssetId, trustLimit?: string) {
+  addAssetSupport(
+    asset: IssuedAssetId,
+    trustLimit?: string,
+  ): SponsoringBuilder {
     super.addAssetSupport(asset, trustLimit);
     return this;
   }
 
-  removeAssetSupport(asset: IssuedAssetId) {
+  removeAssetSupport(asset: IssuedAssetId): SponsoringBuilder {
     super.removeAssetSupport(asset);
     return this;
   }
--- a/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
+++ b/src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
@@ -1,6 +1,7 @@
 import StellarSdk, {
   TransactionBuilder as StellarTransactionBuilder,
   Transaction,
+  xdr,
 } from "stellar-sdk";
 import { IssuedAssetId } from "../../Asset";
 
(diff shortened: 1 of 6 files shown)


src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
```diff
@@ -43,8 +41,17 @@ export class TransactionBuilder {
...
       this.builder.setTimeout(cfg.stellar.defaultTimeout);
     }
+  }
 
-    this.sourceAccount = sourceAccount.accountId();
+  sponsoring(
+    sponsorAccount: AccountKeypair,
+    sponsoredAccount?: AccountKeypair,
+  ): SponsoringBuilder {
+    return new SponsoringBuilder(
```
acharb, line 50: note: slight behind the scenes difference from Kotlin sdk [here](https://github.com/stellar/kotlin-wallet-sdk/blob/ac2a0f1d3944d190a287a22b0c7c9d23a666ee9d/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/transaction/TransactionBuilder.kt#L65-L74).  In Kotlin (if I'm understanding it right) calling .sponsoring returns the same TransactionBuilder instance, after the building function modified the inner operations array.

here we're returning a new SponsoringBuilder instance. This was so instead of the api looking like this:
`txBuilder.sponsoring(sponsorKp, buildingFunction).build()`
(following the lambda function pattern Kotlin is using)

we can do instead:
`txBuilder.sponsoring(sponsorKp).createAccount().build()`

which follows the way the builder looks normally w/o sponsoring

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
```diff
@@ -80,39 +87,24 @@ export class TransactionBuilder {
     return this;
   }
 
-  addOperation(op: xdr.Operation): TransactionBuilder {
-    this.builder.addOperation(op);
+  addAssetSupport(asset: IssuedAssetId, trustLimit?: string) {
```
CassioMG, line 90: could we add return type here?

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
```diff
@@ -80,39 +87,24 @@ export class TransactionBuilder {
...
+    super.addAssetSupport(asset, trustLimit);
     return this;
   }
 
-  setMemo(memo: Memo): TransactionBuilder {
-    this.builder.addMemo(memo);
+  removeAssetSupport(asset: IssuedAssetId) {
```
CassioMG, line 95: could we add return type here?

src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
```diff
@@ -0,0 +1,74 @@
...
+      StellarSdk.Operation.createAccount({
+        destination: newAccount.publicKey,
+        startingBalance: startingBalance.toString(),
+        source: this.sponsorAccount.publicKey,
+      }),
+    );
+    return this;
+  }
+
+  addAssetSupport(asset: IssuedAssetId, trustLimit?: string) {
```
CassioMG, line 40: could we add return type here?

src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
```diff
@@ -0,0 +1,74 @@
...
+    super.addAssetSupport(asset, trustLimit);
+    return this;
+  }
+
+  removeAssetSupport(asset: IssuedAssetId) {
```
CassioMG, line 45: could we add return type here?

CassioMG (approved):
Thanks @acharb!

I left a few comments regarding return types.
Aside from those could we also export the new builder classes on `../Horizon/index.ts` and `src/index.ts` files? So we expose it to wallets - thanks!

src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
Ifropc, line 40: Is it necessary to call super here? Can't we do `return this` in the parent class?

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
Ifropc: How to get back to non-sponsoring builder? 
In kotlin, you can add non-sponsored transactions after 
```kotlin
 stellar
        .transaction(sponsorKeyPair)
        .sponsoring(sponsorKeyPair, newKeyPair) {
          createAccount(newKeyPair)
        }
        .transfer(someOtherAccount) // <- will do a transfer from sponsorKeyPair in the same transaction
```

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
acharb: argh I keep forgetting to do these, thanks for the reminder!
https://github.com/stellar/typescript-wallet-sdk/pull/45/commits/df09183d3e27553519ffc953efc023d7068c407d

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/45/commits/df09183d3e27553519ffc953efc023d7068c407d

src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/45/commits/df09183d3e27553519ffc953efc023d7068c407d

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
acharb: hmm I was under the assumption it would be only a sponsoring transaction or not. How much of a need is it to be able to support allowing both at the same time? 🤔 

A simple solution could be requiring to put non-sponsoring first:
```
 stellar
        .transaction(sponsorKeyPair)
        .transfer(someOtherAccount)
        .sponsoring(sponsorKeyPair, newKeyPair)
        .createAccount()
```


another way, but it would make the code pretty ugly I think, is to do something like:
```
 stellar
        .transaction(sponsorKeyPair)
        .sponsoring(sponsorKeyPair, newKeyPair)
        .createAccount()
        .stopSponsoring()
        .transfer()
```
(would be creating a TransactionBuilder instance, then a Sponsoring one, then a TransactionBuilder one again, unless there's a  cleaner way of doing I'm not thinking of)


I'm in favor of the first option

src/walletSdk/Horizon/Transaction/SponsoringBuilder.ts
acharb: I tried this a bunch last night and for some reason Typescript didn't like it 🤦‍♂️. But yup definitely cleaner this way:
https://github.com/stellar/typescript-wallet-sdk/pull/45/commits/503e00e70e25220539d2e3d7e96a2cee13f7643d

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
Ifropc: First option makes things like this to be impossible 
```kotlin
    val newKeyPair = wallet.stellar().account().createKeyPair()

    val replaceWith = account.createKeyPair()

    val modifyAccountTransaction =
      stellar
        .transaction(sponsoringKey)
        .sponsoring(sponsoringKey, newKeyPair) {
          createAccount(newKeyPair)
          addAccountSigner(replaceWith, 1)
          lockAccountMasterKey()
        }
        .transfer(newKeyPair.address, XLM, "10")
        .build()
        .sign(newKeyPair)
        .sign(sponsoringKey)

    wallet.stellar().submitTransaction(modifyAccountTransaction)
```
Here's [a transaction](https://stellar.expert/explorer/testnet/tx/a325268ff0ae1c220ee876ad5001d8d90206d32250352be91d3c685b62336710) 
Wallet may sponsor creation and modification, and also fund the new account in one transaction with this approach

Ifropc: Can't we accept lambda that takes SponsoringBuilder?
In ts:
```typescript
.sponsoring(sponsoringKey, newKeyPair, spBuilder => {
          spBuilder.createAccount(newKeyPair)
          spBuilder.addAccountSigner(replaceWith, 1)
          spBuilder.lockAccountMasterKey()
        })
```
And internally create SponsoringBuilder, and apply changes from lambda?

acharb: > First option makes things like this to be impossible

hmm ok, in that case I'd vote to do this way:

```
 stellar
        .transaction(sponsorKeyPair)
        .sponsoring(sponsorKeyPair, newKeyPair)
        .createAccount()
        .stopSponsoring()
        .transfer()
```

> Can't we accept lambda that takes SponsoringBuilder?

We can definitely do it that way, but to me it feels counter intuitive

eg. I want to create an account:
`stellar.transaction(sponsorKeyPair).createAccount()`

actually I want to sponsor it:
`stellar.transaction(sponsorKeyPair).sponsoring().createAccount()`

if we do the lambda function way that code will error. I actually implemented it at first using the lambda function way, and made that mistake a couple times so switched it



Maybe it is more intuitive than I'm thinking though. Lmk what you think, I'm happy to do it either way :)

Ifropc: Either option is fine to me, how about we create code samples of 2 alternative options and ask Cassio and Piyal on their opinion how it would be easier to use?

acharb: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1693338265419979

acharb: sgtm for using the lambda function way 👍 
changed here: https://github.com/stellar/typescript-wallet-sdk/pull/45/commits/0a28f7e2c9da40fb7237c3c698ec995f91a06520

Ifropc (approved):

PR #51, opened 2023-09-06
Update README

Ifropc:


README.md
```diff
@@ -1,30 +1,43 @@
...
+let wallet = walletSdk.Wallet.TestNet();
 ```
-yarn test
+
+It should later be re-used across the code, as it has access to various useful children classes. For example, you can 
+authenticate with the `testanchor` as simple as:
+
+```typescript
+const authKey = SigningKeypair.fromSecret("my secret key");
+const sep10 = await anchor.sep10()
```
CassioMG, line 38: Thanks Carl

could we add the line where `testanchor` is created?

I think line 38 is also missing a semicolon at the end

CassioMG (approved):

piyalbasu (approved):

acharb (approved):
lgtm!

PR #52, opened 2023-09-11
only report transactions with changed status

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1088?atlOrigin=eyJpIjoiZWI3YjJkYWZmYmU3NGQ2MDhlNTQ5MjVmYTkzYTM3N2MiLCJwIjoiaiJ9)


src/walletSdk/Watcher/index.ts
```diff
@@ -296,12 +294,13 @@ export class Watcher {
         const registeredTransaction =
           this._transactionsRegistry[assetCode][transaction.id];
 
-        // if we've had the transaction before, only report if there is a change
+        // if we've had the transaction before, only report if there is a status change
+        let isChanged = true;
```
acharb, line 298: in doing this found that the watcher for watching one transaction will stop if we get the same transaction twice

spoke with @CassioMG and agreed it should change to continue watching

CassioMG (approved):
LGTM!

Ifropc (approved):

PR #53, opened 2023-09-12
add default domain signer and default client_domain

acharb:
[default domain signer](https://stellarorg.atlassian.net/browse/WAL-882?atlOrigin=eyJpIjoiMGQ1ZmMzNTBjNTI2NDVlZGI3OWZkZmFmYmRhODdiNzkiLCJwIjoiaiJ9)

[default client_domain](https://stellarorg.atlassian.net/browse/WAL-884?atlOrigin=eyJpIjoiYTRhYjllODM1ZTYxNGFjOTg2OTk0OGM3MWQzYjFjNmUiLCJwIjoiaiJ9)


src/walletSdk/Auth/WalletSigner.ts
```diff
@@ -28,3 +36,36 @@ export const DefaultSigner: WalletSigner = {
...
+      transactionXDR,
+      networkPassphrase,
+    });
+
+    return StellarTransactionBuilder.fromXDR(
+      response.data.transaction,
+      networkPassphrase,
+    ) as Transaction;
+  }
+}
```
JakeUrban, line 71: Is there a way we can require authentication of some form is used here? My concern is that wallets will leave the endpoint that arbitrarily signs transactions unauthenticated because it works out-of-the-box using this function.

acharb: 🤔 we could add something like an authKey param

maybing adding it as an optional field, but just having it there will remind devs they should be thinking about authentication?

eg:

```
async signWithDomainAccount({
    transactionXDR,
    networkPassphrase,
    accountKp,
    authKey?
  }){
 
    const response = await this.client.post(this.url, {
      transactionXDR,
      networkPassphrase,
    }, {
     headers: {
       Authorization: authKey
      }


}
```

any thoughts @Ifropc ?

Ifropc: In Kotlin SDK there is a lambda function parameter left for authentication. It can be bearer, token, etc. 
See[ requestTransformer](https://github.com/stellar/kotlin-wallet-sdk/blob/main/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/auth/WalletSigner.kt#L56)
As for Jake's concern, it's documented, but we can make lambda mandatory, so devs can disable it manually if needed (for test env)

acharb: here you can pass in a custom client (equivalent to the kotlin transformer I believe?). So they could add the header there like this:

```
const customClient = axios.create({
  ...
  headers: {Authorization: 'my-auth-key'}
});

const signer = new DomainSigner(url, customClient)
```

I think in the typescript case requiring creating a custom client would be over kill. In which case I'm in favor of the optional `authKey` field to `signWithDomainAccount`. 

wdyt @JakeUrban

Ifropc: Custom client is not a good idea, cause they need to switch between clients depending on the endpoint.
What if we keep Kotlin pattern?
```ts
const signer = new DomainSigner(url, {Authorization: 'my-auth-key'})
```
(Last argument of client signer is a mandatory headers object) 
So for test env they can do
```ts
const signer = new DomainSigner(url, {})
```

JakeUrban: I'm in favor of the custom client approach, just because `authKey` has the assumption that the server uses the `Authorization` header, when they could use a custom header like `X-Api-Key`.

Ifropc: For custom header they would write
```ts
const signer = new DomainSigner(url, {X-Api-Key: 'my-api-key'})
```

acharb: sounds good to me

just to be clear what did you mean by here? @Ifropc 
```
cause they need to switch between clients depending on the endpoint.
```

by removing the custom client argument the signer will always be using the same `DefaultClient`

Ifropc: Yeah, the thing is that let's say developer define a custom global client with custom timeouts etc.
They would need to not forget to copy all that settings for the domain signer, if the only thing they want to be changed is authentication.
I think it's better if we just append headers to the request, keeping the base client

acharb: ahh got it 👍

acharb: https://github.com/stellar/typescript-wallet-sdk/pull/53/commits/7662ae2545904ba1a55690aea7b4fb429557820a

piyalbasu (approved):

src/walletSdk/Auth/WalletSigner.ts
```diff
@@ -28,3 +37,42 @@ export const DefaultSigner: WalletSigner = {
     );
   },
 };
+
+export class DomainSigner implements WalletSigner {
+  private url: string;
+  private client: AxiosInstance;
+  private headers: HttpHeaders;
+  constructor(url: string, headers: HttpHeaders) {
```
Ifropc, line 45: Can you add a js doc that clarifies that headers is intended to use for authentication?

Ifropc (commented):
LGTM except small comment

src/walletSdk/Auth/WalletSigner.ts
acharb: 👍  https://github.com/stellar/typescript-wallet-sdk/pull/53/commits/02c1922e84ae55360c9fbc248595e21b03098bd1

PR #56, opened 2023-09-13
adding path payment and swap

acharb:
[path payment](https://stellarorg.atlassian.net/browse/WAL-776)
[swap](https://stellarorg.atlassian.net/browse/WAL-777)


src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
```diff
@@ -90,6 +90,35 @@ export class TransactionBuilder extends CommonTransactionBuilder<TransactionBuil
     return this;
   }
 
+  pathPay(
+    destinationAddress: string,
+    sendAsset: StellarAssetId,
+    destAsset: StellarAssetId,
+    amount: string,
+    destMin?: string,
```
acharb, line 98: imo `destMin` can be a bit confusing when first being introduced to path payments. So leaving it optional and using the smallest value possible as default if not given

Ifropc: Can you add doc describing what destMin is? You can copy it from [our documentation ](https://developers.stellar.org/docs/encyclopedia/path-payments)

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
```diff
@@ -90,6 +90,35 @@ export class TransactionBuilder extends CommonTransactionBuilder<TransactionBuil
...
+  ): TransactionBuilder {
+    this.operations.push(
+      StellarSdk.Operation.pathPaymentStrictSend({
```
Ifropc, line 101: Do we not support strictReceive at the time?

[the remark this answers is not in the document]
acharb: kind of similar to `destMin`. If I just want simple case of sending x amount of 1 currency and have receiver receive different currency I don't really care about whether it's `strictReceive` or `strictSend`

I guess it depends on the target audience for this sdk, with have stellar-savy they are

imo I think we should only add a wrapper for strictSend to simplify, and if a developer wants to do strictReceive they can do using `addOperation`

acharb: commented above: https://github.com/stellar/typescript-wallet-sdk/pull/56#discussion_r1324892719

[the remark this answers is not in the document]
acharb: I'm happy to add both methods, lmk if you guys disagree with my thinking
cc @Ifropc @piyalbasu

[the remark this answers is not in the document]
acharb: maybe an optional param flag to `pathPay` to decide between `strictSend` or `strictReceive` is best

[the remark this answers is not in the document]
Ifropc: maybe instead of amount add optional amountSend and amountReceive 
1. Check only one of them specified
2. Do send/receive depending on that amount is provided 
WDYT?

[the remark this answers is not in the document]
acharb: ya I like that plan 👍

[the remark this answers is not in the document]
acharb: updated https://github.com/stellar/typescript-wallet-sdk/pull/56/commits/a897a3fff809af244ff5bdb489de65d898c67a74

src/walletSdk/Types/horizon.ts
```diff
@@ -38,3 +39,13 @@ export enum HORIZON_ORDER {
   ASC = "asc",
   DESC = "desc",
 }
+
+export type PathPayParams = {
+  destinationAddress: string;
+  sendAsset: StellarAssetId;
+  destAsset: StellarAssetId;
+  sendAmount?: string;
```
CassioMG, line 47: should we stick with `string` for amount or try using `number` instead for clients' convenience?

CassioMG: I see we are using `string` for amounts in other parts of the code, maybe it should be a bigger discussion on which type we prefer and agree on a pattern. Wdyt @piyalbasu @Ifropc @acharb ?

CassioMG (approved):
Lgtm! Left a comment regarding `string` vs `number` inputs

src/walletSdk/Types/horizon.ts
CassioMG: I believe most wallets would prefer handling `number` or `BigNumber`, but in case we have a strong reason for keeping `string` that should be fine too. Wallets can always convert the amounts to string whenever necessary

Ifropc: It is a breaking change... We can migrate to numbers in 2.0
(I wanted to create Amount wrapper that stores amount as a number and also an asset id)

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
```diff
@@ -90,6 +95,91 @@ export class TransactionBuilder extends CommonTransactionBuilder<TransactionBuil
...
+  pathPay({
+    destinationAddress,
+    sendAsset,
+    destAsset,
+    sendAmount,
+    destAmount,
+    destMin,
+    sendMax,
+  }: PathPayParams): TransactionBuilder {
+    if ((sendAmount && destAmount) || !(sendAmount || destAmount)) {
```
Ifropc, line 128: [nit] for me (!send && !dest) is a bit more readable

Ifropc (approved):

src/walletSdk/Horizon/Transaction/TransactionBuilder.ts
acharb: 👍  https://github.com/stellar/typescript-wallet-sdk/pull/56/commits/3f35881da6bb54cbaa5dcc9a9d2994f5dd2d7aed

src/walletSdk/Types/horizon.ts
acharb: just to give my 2 cents I like using strings inputs for amounts so I don't have to worry about number overflow or big decimal stuff, and letting the sdk handle those complications

but if their apps are already using number/big number everywhere, then I could see that being annoying having to convert

Ifropc: We should def think it through in the feature and i'm open to discuss it

Ifropc: Though precision is very important, yes. I would like to write something like 
lumens(10) or xlm(5.5) and then pass it where it's needed.

PR #58, opened 2023-09-18
add sep24 example code

acharb:
note: the withdrawal example code will error until release/1.1.3 is merged in


examples/sep24/fullFlow.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const wallet = walletSdk.Wallet.TestNet();
+const stellar = wallet.stellar();
+const anchor = wallet.anchor({ homeDomain: "testanchor.stellar.org" });
+const account = stellar.account();
+
+let kp: SigningKeypair;
+
+const runFullFlow = async () => {
+  const asset = new IssuedAssetId(
```
Ifropc, line 22: Can you use USDC please?

Ifropc (commented):
Can you merge it all in one file please? So it's easier for reader to check just once file (and we can link only this file in the docs)

examples/sep24/fullFlow.ts
acharb: 👍 https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/1222e91f13542b9f296ea0e89f7b70f3f9c46a71

acharb:
> Can you merge it all in one file please? So it's easier for reader to check just once file (and we can link only this file in the docs)

@Ifropc updated. How does it look now?

Ifropc:
Can you follow the kotlin example please? Where it's just one big function that goes over both deposit and withdrawal 
https://github.com/stellar/kotlin-wallet-sdk/blob/main/examples/SEP-24/src/main/kotlin/org/stellar/example/Main.kt

acharb:
ok moved everything into a single file. Is there a reason it needs to be all in one function? Imo it's much easier to read the code when split into methods

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+
+const runSep24 = async () => {
+  const asset = new IssuedAssetId(
+    "USDC",
+    "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5",
+  );
+
+  // Create Account
+
+  const createAccount = async () => {
```
Ifropc, line 28: Can you remove it out of runSep24 function please?

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+    console.log("creating account ...");
+    kp = account.createKeypair();
+    console.log(`kp: \n${kp.publicKey}\n${kp.secretKey}`);
+    await axios.get("https://friendbot.stellar.org/?addr=" + kp.publicKey);
```
Ifropc, line 32: Can you add a comment that this is for funding the new account

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+  });
+
+  stop = resp.stop;
+};
+
+// Create Withdrawal
+let authTokenWithdraw;
+export const runWithdraw = async (anchor, kp) => {
+  console.log("\ncreating withdrawal ...");
+  const auth = await anchor.sep10();
```
Ifropc, line 122: It's already authenticated above, can we reuse token?

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+  };
+
+  const onError = (error: Types.AnchorTransaction | Error) => {
+    console.log({ error });
+  };
+
+  const watcher = anchor.sep24().watcher();
+  const resp = watcher.watchAllTransactions({
+    authToken: authTokenDeposit,
+    assetCode: "USDC",
```
Ifropc, line 108: IMO we should define asset as a constant once, and then get asset code as `asset.getCode()`

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// Watch Deposit
+export let depositDone = false;
+export const runDepositWatcher = (anchor: Anchor) => {
+  console.log("\nstarting watcher ...");
+
+  let stop: Types.WatcherStopFunction;
+  const onMessage = (m: Types.AnchorTransaction) => {
+    console.log({ m });
+    if (m.status === "completed") {
```
Ifropc, line 94: Don't we have enum for statuses?

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+  const auth = await anchor.sep10();
+  authTokenDeposit = await auth.authenticate({ accountKp: kp });
+
+  const assetCode = "USDC";
+  const resp = await anchor.sep24().deposit({
+    assetCode,
+    authToken: authTokenDeposit,
+    lang: "en-US",
+    destinationMemo: new Memo(MemoText, "test-memo"),
+    extraFields: {
```
Ifropc, line 77: This 2 fields are deprecated in the standard, let's not use them in the example

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  const txBuilder = await stellar.transaction({
+    sourceAddress: kp,
+    baseFee: 1000,
+  });
+  const tx = txBuilder
+    .transferWithdrawalTransaction(withdrawalTxn, asset)
+    .build();
+  kp.sign(tx);
+  const success = await stellar.submitTransaction(tx);
```
Ifropc, line 152: This method always returns true, you don't need to check for the result being returned (we should make this method to return `promise<null`  btw;

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+    console.log({ m });
+
+    if (m.status === "pending_user_transfer_start") {
+      sendWithdrawalTransaction(m, kp);
+    }
+
+    if (m.status === "completed") {
+      console.log("status completed, stopping watcher");
+
+      stop();
```
Ifropc, line 173: what does `stop` do?

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+      process.exit(0);
```
Ifropc, line 174: why do we need this?

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,207 @@
...
+    }
+  };
+
+  const onError = (e) => {
+    console.log({ e });
```
Ifropc, line 179: console.error?

examples/sep24/sep24.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: good point https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: 👍  https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: it stops the watcher from watching

examples/sep24/sep24.ts
acharb: ya we don't good point https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3d0558484d2e6ee2a54b2ed2836c999995ef28de

examples/sep24/sep24.ts
Ifropc: Where do we get it from?

examples/sep24/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "The sdk can be found [here](https://github.com/stellar/typescript-wallet-sdk).",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "ts-node sep24.ts"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@stellar/typescript-wallet-sdk": "^1.1.2",
```
Ifropc, line 13: Won't it pull this version from npm?

examples/sep24/sep24.ts
```diff
@@ -0,0 +1,189 @@
...
+  "USDC",
+  "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5",
+);
+
+const runSep24 = async () => {
+  await createAccount();
+  await runDeposit(anchor, kp);
+  await runDepositWatcher(anchor);
+
+  while (!depositDone) {
```
Ifropc, line 30: Do we not need to do it for withdrawal too?

Ifropc: Also, I think it should be inside runDepositWatcher

acharb: > Do we not need to do it for withdrawal too?

no the watcher keeps watching until stop() is called, which we only call when the status is returned as completed

> Also, I think it should be inside runDepositWatcher

similar the watcher for the deposit will also keep running until its stop() is called. The only reason there's a while loop is so we can listen for the deposit watcher to finish before asking to start the withdrawal flow. Imo it's a separate idea, since it's just there to connect the deposit to the withdrawal (and not needed unless you're running code like this)

examples/sep24/package.json
acharb: ya, if you're thinking of pulling the source code from under the src folder I'm not sure you can do that 🤔  or if it's a good idea for separation of concerns reason

I can look into this tomorrow

acharb: imo it's fine separate, in it's own sub-module. It's use is essentially for documentation

examples/sep24/sep24.ts
Ifropc: > no the watcher keeps watching until stop() is called, which we only call when the status is returned as completed

Won't the main thread die if we don't do the while loop after withdrawal flow?

examples/sep24/package.json
Ifropc: I think it's a good idea just because the code is always up-to-date...

Ifropc: Yeah, it should be separate, but I'm curious how easy it is to get dependency from another folder. I was never been able to do it clean, to be honest

examples/sep24/sep24.ts
acharb: > Won't the main thread die if we don't do the while loop after withdrawal flow?

no the watcher will keep running until `stop` is called. See here: https://github.com/stellar/typescript-wallet-sdk/blob/8a7d22f432b23ad62b743486509d5394e548d155/src/walletSdk/Watcher/index.ts#L195-L205

once the call to the anchor is made, it waits for timeout length and then calls again it (and so on)

examples/sep24/package.json
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/58/commits/3874415277787e7f71b294edb7f5fe5667aa8b79
looks to be easier than I thought lol. This set up should work. Doesn't affect src build or test, which is what I was worried about.

though note: it won't break the sdk build or anything if out-of-date with the source code. We can add a script or linting-type thing later though to check it

Ifropc (approved):

PR #60, opened 2023-09-19
Account setup and recovery using SEP-30

CassioMG:
https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?selectedIssue=WAL-763

This mimics the Kotlin implementation for [account recovery](https://github.com/stellar/kotlin-wallet-sdk/tree/main/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/recovery)

I'll try adding more tests as it has only a few at the moment


acharb:
on first glance looks good. I'll be able to review more in depth beginning of next week if others don't get to first

I think we should make sure we have tests (w/ mocked responses if needed) for the class methods though. eg `deduceKey`, `replaceDeviceKey`, `createRecoverableWallet`, `getAccountInfo` etc

piyalbasu:
@CassioMG Ah, sorry, I closed the original release/1.2.0 branch. Would you mind reopening this on release/1.3.0? Thanks!

PR #66, opened 2023-09-25
add type aliases, and make sep24 example use env vars

acharb:
[type aliases](https://stellarorg.atlassian.net/browse/WAL-1102)

[use env vars in sep24 example](https://stellarorg.atlassian.net/browse/WAL-1126)


examples/sep24/sep24.ts
```diff
@@ -182,6 +215,27 @@ export const runWithdrawWatcher = (anchor, kp) => {
...
+
+  const signer = SigningKeypair.fromSecret(clientSecret);
+
+  const transaction = TransactionBuilder.fromXDR(
+    transactionXDR,
+    networkPassphrase,
+  ) as Transaction;
+  signer.sign(transaction);
+
+  return Promise.resolve(transaction);
```
piyalbasu, line 236: just curious - what's the benefit of making this function `async`? Everything within this function runs synchronously and then we return a resolved Promise. We're not actually waiting for anything async to resolve

piyalbasu (approved):

examples/sep24/sep24.ts
acharb: ah good question. It's because in practice the `signWithDomainAccount` method is for signing the transaction by a backend server, see [here](https://github.com/stellar/typescript-wallet-sdk/blob/a97ea11976d1b08bd47fb2c46b562e3a2667b17d/src/walletSdk/Auth/WalletSigner.ts#L74-L86)
(or at least giving the user the option to sign it w/ a server)

but here we're signing with a .env secret for the example, so doesn't have to be async but making it so to follow the function type

PR #69, opened 2023-10-04
better axios error handling (pt 2)

acharb:
in making the changes from the PR comments here: https://github.com/stellar/typescript-wallet-sdk/pull/67

I decided to return the error with `AxiosErrorData` in the data field instead of one large string, which will make using the error object much easier if needed

the error in total will look like this if a http request fails:
Screen Shot 2023-10-03 at 9 26 00 PM

### DIFF
--- a/src/walletSdk/Exceptions/index.ts
+++ b/src/walletSdk/Exceptions/index.ts
@@ -1,12 +1,15 @@
 import { Networks, Horizon } from "stellar-sdk";
 import axios from "axios";
-import { AnchorTransaction, FLOW_TYPE } from "../Types";
+import { AnchorTransaction, FLOW_TYPE, AxiosErrorData } from "../Types";
 import { extractAxiosErrorData } from "../Utils";
 
 export class ServerRequestFailedError extends Error {
+  data: AxiosErrorData;
+
   constructor(e: Error) {
     if (axios.isAxiosError(e)) {
-      super(`Server request failed with error: ${extractAxiosErrorData(e)}`);
+      super(`Server request failed with error`);
+      this.data = extractAxiosErrorData(e);
     } else {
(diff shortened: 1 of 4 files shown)


src/walletSdk/Exceptions/index.ts
```diff
@@ -1,12 +1,15 @@
...
 import { extractAxiosErrorData } from "../Utils";
 
 export class ServerRequestFailedError extends Error {
+  data: AxiosErrorData;
+
   constructor(e: Error) {
     if (axios.isAxiosError(e)) {
-      super(`Server request failed with error: ${extractAxiosErrorData(e)}`);
+      super(`Server request failed with error`);
+      this.data = extractAxiosErrorData(e);
```
CassioMG, line 12: I like this more structured approach, I think it helps a lot codewise to have a data object to properly handle the error.

But, I think it's also useful to have a hint of the error at a glance, and also for error tracking tools like Sentry I believe it would pretty much display only the "Server request failed with error" text on the main error list, which means we would need to open the individual error to only then be able to take a closer look at the data values and understand what this error is about.

What do you think about making something like this? 
```
this.data = extractAxiosErrorData(e);
super(`Server request failed with error: ${this.data.status} ${this.data.responseData.toString() ?? this.data.statusText}`);
```
With this I believe we'd have the best of both worlds: a proper `data` structure to handle the error in the code, and a more descriptive main error message to speed-up human diagnostics.

acharb: I like that idea a lot 👍  https://github.com/stellar/typescript-wallet-sdk/pull/69/commits/27460627af0fa8c0b5a8abb5803adbd66df8c47e

CassioMG (approved):
Lgtm!

PR #71, opened 2023-10-09
add Customer / SEP-12 code

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-866)


README.md
```diff
@@ -24,21 +27,25 @@ npm install @stellar/typescript-wallet-sdk
...
 const anchor = wallet.anchor({ homeDomain: "testanchor.stellar.org" });
 const sep10 = await anchor.sep10();
 
-const authToken = await sep10.authenticate({accountKp: authKey});
+const authToken = await sep10.authenticate({ accountKp: authKey });
 ```
 
-Read [full wallet guide](https://developers.stellar.org/docs/category/build-a-wallet) for more info
+Read
+[full wallet guide](https://developers.stellar.org/docs/category/build-a-wallet-with-the-wallet-sdk)
```
acharb, line 50: just updating the URL here, rest of changes were latent formatting fixes

src/walletSdk/Types/sep12.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export enum Sep12Type {
+  string = "string",
+  binary = "binary",
+  number = "number",
+  date = "date",
+}
+
+export type Field = {
+  type: Sep12Type;
```
acharb, line 21: @Ifropc kotlin has a couple of these fields optional (eg. https://github.com/stellar/kotlin-wallet-sdk/blob/3b81495e021939153d7f713a666b081ebbe6a61c/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/customer/Data.kt#L21-L22)

according to the [SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#fields) they should be required, lmk if there was a reason to make them optional though and I can change these

test/customer.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    const anchor = wallet.anchor({ homeDomain: "testanchor.stellar.org" });
+
+    const auth = await anchor.sep10();
+    const authToken = await auth.authenticate({ accountKp });
+
+    const sep12 = await anchor.sep12(authToken);
+    const customerType = "sep31-receiver";
+    const customerEmail = `${getRandomString(6)}@gmail.com`;
+
+    // Update
```
CassioMG, line 31: Should the comment say "Add" instead of "Updated"? Since we have an "Update" on line 58

src/walletSdk/Customer/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
+    );
+
+    if (!resp.data.id) {
+      throw new CustomerNotFoundError(id);
+    }
+    return resp;
+  }
+
+  async add(
```
CassioMG, line 43: @acharb @Ifropc I am missing a way of properly handling `binary` files. According to the SEP, we could either upload it through [PUT /customer](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#request-1) as the last fields. Or use the new [POST /customer/files](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#post-request) endpoint. In both cases I believe the [Content-Type](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#content-type) header should be set to `multipart/form-data` which I am not finding in the current TS implementation (didn't check the Kotlin one)

Ifropc: It's not supported in KT at the moment. I think we should support it eventually in TS

src/walletSdk/Customer/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.token = token;
+    this.baseUrl = baseUrl;
+    this.httpClient = httpClient;
+    this.headers = {
+      "Content-Type": "application/json",
+      Authorization: `Bearer ${this.token}`,
+    };
+  }
+
+  async getByIdAndType(id: string, type: string): Promise<GetCustomerResponse> {
```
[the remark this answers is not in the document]
Ifropc: I think ideally you should provide either id or type. I know there is an edge case, where there is only one type, so specifying it is not necessary, but I think we should make it clear to the user that one of this fields should be provided.
I checked kt implementation -> you are right, it should be fixed there

src/walletSdk/Types/sep12.ts
Ifropc: I don't remember why they were left optional. It was either by mistake, or because testanchor didn't provide them for some reason

Ifropc: If it's always provided by the testanchor I don't see why we shouldn't make them mandatory

src/walletSdk/Customer/index.ts
Ifropc, line 29: Maybe we can split this to `getById` and `getByType`? The get method also accept optional parameters, which were not supported by Kotlin for some reason.

test/customer.test.ts
acharb: yes ty https://github.com/stellar/typescript-wallet-sdk/pull/71/commits/e3f8ab3658106304826194b4c61f49b3ee6f7dc6

src/walletSdk/Anchor/index.ts
```diff
@@ -89,6 +95,30 @@ export class Anchor {
     return this.sep10();
   }
 
+  /**
+   * Create new customer object to handle customer records with the anchor using SEP-12.
+   * @param {string} authToken - The authentication token.
+   * @returns {Promise<Sep12>} - A Promise that resolves to a Sep12 instance.
+   * @throws {KYCServerNotFoundError} - If the KYC server information is not available.
+   */
+  async sep12(authToken: string): Promise<Sep12> {
```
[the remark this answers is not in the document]
acharb: https://stellarorg.atlassian.net/browse/WAL-1152

src/walletSdk/Customer/index.ts
acharb: I don't think there's a case where you can GET with just id and no type right? According to sep:
```
GET [KYC_SERVER]/customer&type=<customer-type>
GET [KYC_SERVER]/customer?account=<stellar-account>&type=<customer-type>
GET [KYC_SERVER]/customer?account=<stellar-account>&memo=<memo>&type=<customer-type>
GET [KYC_SERVER]/customer?account=<muxed-account>&type=<customer-type>
GET [KYC_SERVER]/customer?id=<customer-id>&type=<customer-type>
```

what about just `getCustomer` with optional id, type, and account?

and in a jsdoc we explain what combos are needed and link to the [SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get)

CassioMG:
```
GET [KYC_SERVER]/customer&type=<customer-type>
GET [KYC_SERVER]/customer?account=<stellar-account>&type=<customer-type>
GET [KYC_SERVER]/customer?account=<stellar-account>&memo=<memo>&type=<customer-type>
GET [KYC_SERVER]/customer?account=<muxed-account>&type=<customer-type>
GET [KYC_SERVER]/customer?id=<customer-id>&type=<customer-type>
```
I thought those were a few samples to illustrate some possible variations but not all of them, but I'm also not sure about it

I like the idea of a single `getCustomer` function with optional fields

CassioMG: > what about just getCustomer with optional id, optional type, and optional account?

what if we leave only `id`, `type` and `memo` as optional params? since the other optional fields (like `account`) are deprecated

acharb: how does this look? https://github.com/stellar/typescript-wallet-sdk/pull/71/commits/af6813b1a96eef68ecf70d638d06df74915b58d6

src/walletSdk/Customer/index.ts
```diff
@@ -26,14 +31,36 @@ export class Sep12 {
...
+    if (id) {
+      queryParams.push(`id=${id}`);
+    }
+    if (memo) {
+      queryParams.push(`memo=${memo}`);
+    }
+    if (lang) {
+      queryParams.push(`lang=${lang}`);
+    }
+    const queryString = queryParams.join("&");
```
CassioMG, line 70: @acharb hey, so the `queryString` lib has this very cool method which we could use to replace this chunk of code in case you'd like to take a look.

You could simply do something like this:

```
import queryString from "query-string";

async getCustomer(params: GetCustomerParams): Promise<GetCustomerResponse> {

   const qs = queryString.stringify(params);
   ...
   `${this.baseUrl}/customer?${qs}`
   ...
}
```

src/walletSdk/Customer/index.ts
```diff
@@ -26,14 +31,36 @@ export class Sep12 {
...
   }
 
-  async getByIdAndType(id: string, type: string): Promise<GetCustomerResponse> {
+  async getCustomer({
+    id,
+    type,
+    memo,
+    lang,
+  }: GetCustomerParams): Promise<GetCustomerResponse> {
+    if (!id && !type && !memo) {
```
CassioMG, line 54: @acharb @Ifropc  I'm not sure we should be enforcing any of those params, I could see Anchors being able to detect clients by its Sep-10 auth token. What do you think?

src/walletSdk/Customer/index.ts
CassioMG: looks great! Left a couple comments/suggestions

src/walletSdk/Customer/index.ts
acharb: good point. Adjusted it here wdyt?
https://github.com/stellar/typescript-wallet-sdk/pull/71/commits/85e5213c54d74270beb0b64b93e4474e81f9ace1

src/walletSdk/Customer/index.ts
acharb: ahh nice I didn't know we had query-string installed already much cleaner
https://github.com/stellar/typescript-wallet-sdk/pull/71/commits/f58a1895a258b6c115e1beec6601f30e80f284e4

src/walletSdk/Customer/index.ts
acharb: ya I agree on second thought, removed:

https://github.com/stellar/typescript-wallet-sdk/pull/71/commits/f58a1895a258b6c115e1beec6601f30e80f284e4

src/walletSdk/Customer/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  Field,
+  ProvidedField,
+  GetCustomerParams,
+  GetCustomerResponse,
+  AddCustomerResponse,
+} from "../Types";
+
+// Used for identifying binary fields.
+// https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0009.md
+const Sep9BinaryFields = [
```
CassioMG, line 17: @acharb so we have those fields suggested by Sep-9 but Anchor and wallets could add any other "random" fields they would like to. This means if they decide on sending let's say a `photo_self` they wouldn't be able to send it as a `binary` if we limit it to what's only included in `Sep9BinaryFields`.

A more generic/flexible approach would be augmenting the `CustomerInfoMap` to account for field types.

Something like:
```
export type CustomerInfoMap = {
  [field_key: string]: {
     type: Sep12Type;
     value: any;
   }
};
```

So we could detect if we have any `binary` field being passed to set the "multipart/form-data" and append those `binary` fields as the last fields in the request body.

What do you think?

CassioMG: hummm maybe a much simpler approach would be just adding a `binaries` array to the add/update function input in which the client could list the field keys which are `binary`.

src/walletSdk/Customer/index.ts
Ifropc: True

src/walletSdk/Customer/index.ts
acharb: I like the idea of adding a binary field, but I was thinking the binary map similar to the sep9Info.

I like this idea as well so that the user doesn't have to remember that binary fields need to be in the map last

lmk what you think

https://github.com/stellar/typescript-wallet-sdk/pull/71/commits/263cbaa3b6740ca0bc6c691aa980843dc70d6990

CassioMG: love it!

CassioMG (approved):

src/walletSdk/Customer/index.ts
acharb: thanks for the help  🙌

PR #75, opened 2023-10-30
small fixes to Recovery code

acharb:
some fixes to the recovery code found while running integration tests

### DIFF
--- a/test/transaction.test.ts
+++ b/test/transaction.test.ts
@@ -291,9 +291,7 @@ describe("Path Payment", () => {
         sendAmount: "5",
       })
       .build();
-    sourceKp.sign(txn);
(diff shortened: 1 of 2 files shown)


test/transaction.test.ts
```diff
@@ -291,9 +291,7 @@ describe("Path Payment", () => {
         sendAmount: "5",
       })
       .build();
-    sourceKp.sign(txn);
```
acharb, line 294: unrelated fix. This test this fails intermintedly when usdc-xlm order book is volatile on testnet, so fixing to just check  the operation type (.swap is still submitting to the network and uses .pathPay)

Ifropc (approved):

CassioMG (commented):
Thanks for fixing those!

PR #77, opened 2023-11-07
add SEP6 deposit and withdrawal

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1059)

### DIFF
--- a/src/walletSdk/Anchor/Sep6.ts
+++ b/src/walletSdk/Anchor/Sep6.ts
@@ -69,8 +69,8 @@ export class Sep6 {
    * @param {string} options.authToken - The authentication token.
    * @param {Sep6DepositParams} options.params - The parameters for the deposit request.
    *
-   * @returns {Promise<Sep6DepositResponse>} Sep6 deposit response, containing next steps to complete
-   * the deposit.
+   * @returns {Promise<Sep6DepositResponse>} Sep6 deposit response, containing next steps if needed
+   * to complete the deposit.
    *
    * @throws {Sep12MissingInfoError} If Sep-12 KYC info is missing for the user.
    * @throws {Sep6DepositDeniedError} If the deposit is still being processed or not accepted.
--- a/src/walletSdk/Anchor/Sep6.ts
+++ b/src/walletSdk/Anchor/Sep6.ts
@@ -1,13 +1,20 @@
 import { AxiosInstance } from "axios";
+import queryString from "query-string";
 
 import { Anchor } from "../Anchor";
-import { ServerRequestFailedError } from "../Exceptions";
-import { Sep6Info } from "../Types";
-
-type Sep6Params = {
-  anchor: Anchor;
-  httpClient: AxiosInstance;
-};
+import {
+  ServerRequestFailedError,
+  Sep12MissingInfoError,
+  Sep6DepositDeniedError,
+} from "../Exceptions";
+import {
+  Sep6Info,
+  Sep6Params,
+  Sep6DepositParams,
+  Sep6WithdrawParams,
+  Sep6DepositResponse,
+  Sep6WithdrawResponse,
+} from "../Types";
 
 /**
  * Flow for creating deposits and withdrawals with an anchor using SEP-6.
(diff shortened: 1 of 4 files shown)


src/walletSdk/Anchor/Sep6.ts
```diff
@@ -53,4 +60,79 @@ export class Sep6 {
...
+          headers: {
+            "Content-Type": "application/json",
+            Authorization: `Bearer ${authToken}`,
+          },
+        },
+      );
+      return resp.data;
+    } catch (e) {
+      if (e.response?.data?.type === "non_interactive_customer_info_needed") {
+        throw new Sep12MissingInfoError(e.response?.data?.fields);
```
Ifropc, line 101: I don't think it's an error. It's expected behavior in the flow. I think we should have different types of the Sep6Response which wallets must handle accordingly

src/walletSdk/Anchor/Sep6.ts
```diff
@@ -53,4 +60,79 @@ export class Sep6 {
...
+   * @param {Sep6DepositParams} options.params - The parameters for the deposit request.
+   *
+   * @returns {Promise<Sep6DepositResponse>} Sep6 deposit response, containing next steps if needed
+   * to complete the deposit.
+   *
+   * @throws {Sep12MissingInfoError} If Sep-12 KYC info is missing for the user.
+   * @throws {Sep6DepositDeniedError} If the deposit is still being processed or not accepted.
+   * @throws {Error} If an unexpected error occurs during the deposit operation.
+   */
+  async deposit({
```
Ifropc, line 79: What about exchange endoints, are they out of scope?

src/walletSdk/Anchor/Sep6.ts
Ifropc: BTW you can find flow diagrams for SEP-6 here
https://github.com/stellar/java-stellar-anchor-sdk/tree/develop/docs/diagrams/sep6

Ifropc (approved):

src/walletSdk/Anchor/Sep6.ts
acharb: good point, how about this? https://github.com/stellar/typescript-wallet-sdk/pull/77/commits/fb25d34c7e844d11cb12ac976eedb77ec9dd441c

also I realized those resposnes can be for deposit or withdrawal so i added a helper flow function similar to sep24

src/walletSdk/Anchor/Sep6.ts
acharb: ya I thought that's what this ticket is for right?

https://stellarorg.atlassian.net/browse/WAL-1060?atlOrigin=eyJpIjoiNjYzZmI4ZWExYjI3NGVmMjljYTgwYmI4MjIwMjUzZWQiLCJwIjoiaiJ9

src/walletSdk/Anchor/Sep6.ts
```diff
@@ -53,4 +57,87 @@ export class Sep6 {
...
+        },
+      );
+      return {
+        type: Sep6ResponseType.SUCCESS,
+        data: resp.data,
+      };
+    } catch (e) {
+      if (e.response?.data?.type === "non_interactive_customer_info_needed") {
+        return {
+          type: Sep6ResponseType.MISSING_KYC,
```
Ifropc, line 131: Do we need to explicitly provide type to all this responses?

acharb: it makes it a bit clearer for the user to handle with typescript 

eg. w/ a type in the response:
```
  const resp = deposit(...)
  if (resp.type === Sep6ResponseType.MISSING_KYC) {
    console.log((resp.data as Types.Sep6MissingKYC).fields);
  }
```

otherwise have to do something like:
```
if ("fields" in resp) {
   console.log((resp.data as Types.Sep6MissingKYC).fields);
  }
```

I'm not tied to it, I can remove it. Then the response will look like this:
```
export interface Sep6DepositResponse {
  data: Sep6DepositSuccess | Sep6MissingKYC | Sep6Pending;
}
```

acharb: after thinking about it more I think you're right removing the type field and using a Union type is best, and is the more standard typescript way of handling

https://github.com/stellar/typescript-wallet-sdk/pull/77/commits/91b6268809501cc6b710d7abf4c874f1801c27ba

Ifropc (approved):

PR #79, opened 2023-11-28
add polling to sep6

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1178?atlOrigin=eyJpIjoiMTUzMmE4ZTMyZmFjNDlhMGIxMzUwYTM0M2I3ZjVmYjIiLCJwIjoiaiJ9)

re-using a lot of the same logic already there for polling sep24 transactions

### DIFF
--- a/src/walletSdk/Anchor/Sep24.ts
+++ b/src/walletSdk/Anchor/Sep24.ts
@@ -1,13 +1,10 @@
 import { AxiosInstance } from "axios";
-import queryString from "query-string";
 
 import { Anchor } from "../Anchor";
 import {
   AssetNotSupportedError,
   ServerRequestFailedError,
   MissingTransactionIdError,
-  InvalidTransactionResponseError,
-  InvalidTransactionsResponseError,
 } from "../Exceptions";
 import {
   FLOW_TYPE,
@@ -17,8 +14,13 @@ import {
   GetTransactionParams,
   GetTransactionsParams,
   AnchorServiceInfo,
+  WatcherSepType,
 } from "../Types";
-import { Watcher } from "../Watcher";
+import {
+  Watcher,
+  _getTransactionsForAsset,
+  _getTransactionBy,
+} from "../Watcher";
 import { camelToSnakeCaseObject } from "../Utils";
 
 // Let's prevent exporting this constructor type as
@@ -182,7 +184,7 @@ export class Sep24 {
    * @returns {Watcher} A new Watcher instance.
    */
   watcher(): Watcher {
-    return new Watcher(this.anchor);
+    return new Watcher(this.anchor, WatcherSepType.SEP24);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/walletSdk/Anchor/Sep24.ts
```diff
@@ -212,38 +214,20 @@ export class Sep24 {
     const toml = await this.anchor.sep1();
     const transferServerEndpoint = toml.transferServerSep24;
 
-    let qs: { [name: string]: string } = {};
```
acharb, line 215: basically just moving this code to a helper function so can be used for both seps

src/walletSdk/Types/anchor.ts
```diff
@@ -37,6 +37,7 @@ export interface ProcessingAnchorTransaction extends BaseTransaction {
   amount_out_asset?: string;
   amount_out: string;
   amount_fee_asset?: string;
+  quote_id?: string;
```
acharb, line 40: note sure why this was left off (I may have just missed it before)

CassioMG (approved):

src/walletSdk/Anchor/Sep6.ts
```diff
@@ -177,4 +190,104 @@ export class Sep6 {
...
+   * @param {string} [params.pagingId] - The response should contain transactions starting prior to this ID (exclusive).
+   * @param {string} [params.lang] - The desired language (localization), it can also accept locale in the format 'en-US'.
+   * @returns {Promise<Sep6Transaction[]>} A list of transactions as requested by the client, sorted in time-descending order.
+   * @throws {InvalidTransactionsResponseError} Anchor returns an invalid response.
+   * @throws {ServerRequestFailedError} If server request fails.
+   */
+  async getTransactionsForAsset({
+    authToken,
+    assetCode,
+    account,
```
Ifropc, line 222: Why do we need an account here? Can't we get it from the auth token? I think in real implementations there are no use cases where account != SEP-10 account

acharb: Sep 6 requires it. If you don't give an account that matches the auth token account it gives a 403. 

I'm not with my laptop right now but I can link to it on the sep 6 spec tonight

Ifropc: I'm suggesting to omit it in the method parameter, and instead always pass account from the auth token to the anchor. (that's how anchor platform and polaris work, as they require authentication in /transaction endpoint, therefore account must always match authenticated account)

acharb: ahh I see. Ya I agree that's better. We have this ticket for tying the account to the auth token (right now it's just a string) https://stellarorg.atlassian.net/browse/WAL-1152?atlOrigin=eyJpIjoiYjMxYWY2NDRkY2IyNGE4MGE4MjlkNWM3NDFmNzVlMDQiLCJwIjoiaiJ9

I'll make a note on the ticket to change this as well to always pull from the auth token

PR #81, opened 2023-12-11
add authToken class

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1152?atlOrigin=eyJpIjoiYmFkMjdjOWRhNDE4NDY1ZGE5OTY0ODhmMzE4MmE5NzEiLCJwIjoiaiJ9)

changes authToken to be an object for better usage

### DIFF
--- a/examples/sep24/sep24.ts
+++ b/examples/sep24/sep24.ts
@@ -88,7 +88,7 @@ const createAccount = async () => {
 };
 
 // Create Deposit
-let authToken: string;
+let authToken: Types.AuthToken;
 export const runDeposit = async (anchor: Anchor, kp: SigningKeypair) => {
   console.log("\ncreating deposit ...");
   const auth = await anchor.sep10();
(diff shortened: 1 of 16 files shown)


acharb:
@Ifropc I was planning on changing the code [here](https://github.com/stellar/typescript-wallet-sdk/pull/79#discussion_r1410092438) to not include account as an arg like we talked about, and just always pull from the `authToken`.

Though I was thinking if an anchor doesn't use sep10 auth then they'd need an account arg. Do you think it's worth considering that case?

acharb:
@Ifropc  ok cool, removed the unneeded account arg https://github.com/stellar/typescript-wallet-sdk/pull/81/commits/61cec1e6c87f322d0553c170454e489a664048f5

also can you approve here? 🙏 small update to the docs:
https://github.com/stellar/stellar-docs/pull/285

Ifropc (approved):

PR #94, opened 2024-01-18
first pass at adding wallet interface and soroban signing

piyalbasu:


src/walletSdk/Wallet/index.ts
```diff
@@ -0,0 +1,37 @@
+import FreighterApi from "@stellar/freighter-api";
```
willemneal, line 1: Perhaps it's best if we flipped this. This way the frieghter API package could provide the Wallet implementation.

piyalbasu: Ah yeah, that makes sense. The idea being that the freighter-api is the default interface that we expect from a wallet.

For added context: We're also talking about creating some helper methods in this lib to allow other browser extensions to quickly implement this expected interface in their own codebase so they can quickly make use of `signAndSend` and `signAuthEntries`

package.json
```diff
@@ -41,7 +41,8 @@
     "webpack-cli": "^5.1.1"
   },
   "dependencies": {
-    "@stellar/stellar-sdk": "^11.1.0",
+    "@stellar/freighter-api": "^1.7.1",
+    "@stellar/stellar-sdk": "stellar/js-stellar-sdk#bindings",
```
CassioMG, line 45: @piyalbasu just out of curiosity, what's the reason behind this "#bindings" change?

piyalbasu: This was targeted at a branch of stellar-sdk that had the TS bindings update for dev purposes. This work may be merged already. I'll follow up

PR #95, opened 2024-01-23
add sep10 sign challenge txn helper

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1235?atlOrigin=eyJpIjoiMzNlYjg0NThkYTcyNDY2ZWIyNjEwNzBiMzgzZGQ4YWQiLCJwIjoiaiJ9)

validates
- sequence num is 0
- signed by anchor

then returns signed txn

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,7 +16,12 @@ export {
 export { Anchor } from "./walletSdk/Anchor";
 export { Sep24 } from "./walletSdk/Anchor/Sep24";
 export { IssuedAssetId, NativeAssetId, FiatAssetId } from "./walletSdk/Asset";
-export { Sep10, WalletSigner, DefaultSigner } from "./walletSdk/Auth";
+export {
+  Sep10,
+  WalletSigner,
+  DefaultSigner,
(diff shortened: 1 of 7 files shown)


piyalbasu (approved):

Ifropc:
This is for server side, right? Can we move it to a different place so developer won't confuse it for client side signing?

acharb:
> This is for server side, right? Can we move it to a different place so developer won't confuse it for client side signing?

I left it in the main walletSdk because it _can_ be used for either client or server side right?

Ifropc:
I think it's technically only for server, because it's used to sign client domain challenge (which must only be on the server)

acharb:
@Ifropc  gotchya, hows this then?
https://github.com/stellar/typescript-wallet-sdk/pull/95/commits/2b7cfbbf4d9205e56ca8d1494750880d10eec694

acharb:
hmm we already have a `Utils` module at that level, having two Utils at same level might be a little repetitive?

Will merge this in though but open to changing the name

PR #97, opened 2024-02-05
add sep24 callback params

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1280?atlOrigin=eyJpIjoiYjQ4YWE2MmMyNGFlNGUzNTgzMzczYjQ1NGM2MjliNmQiLCJwIjoiaiJ9)

### DIFF
--- a/src/walletSdk/Anchor/Sep24.ts
+++ b/src/walletSdk/Anchor/Sep24.ts
@@ -71,6 +71,8 @@ export class Sep24 {
     extraFields,
     destinationMemo,
     destinationAccount,
+    callback,
+    on_change_callback,
   }: Sep24PostParams): Promise<Sep24PostResponse> {
     return this.flow({
       assetCode,
@@ -79,6 +81,8 @@ export class Sep24 {
       extraFields,
       destinationMemo,
       account: destinationAccount,
+      callback,
+      on_change_callback,
       type: FLOW_TYPE.DEPOSIT,
     });
   }
@@ -100,6 +104,8 @@ export class Sep24 {
     lang,
     extraFields,
     withdrawalAccount,
+    callback,
+    on_change_callback,
   }: Sep24PostParams): Promise<Sep24PostResponse> {
     return this.flow({
       assetCode,
@@ -107,6 +113,8 @@ export class Sep24 {
       lang,
       extraFields,
       account: withdrawalAccount,
+      callback,
+      on_change_callback,
       type: FLOW_TYPE.WITHDRAW,
     });
   }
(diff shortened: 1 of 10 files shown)


src/walletSdk/Anchor/Sep24.ts
```diff
@@ -61,6 +61,10 @@ export class Sep24 {
    * @param {ExtraFields} [params.extraFields] - Additional fields for the request.
    * @param {Memo} [params.destinationMemo] - Memo information for the destination account.
    * @param {string} [params.destinationAccount] - The destination account for the deposit.
+   * @param {string} [params.callback] - The callback URL or postMessage the anchor should POST to
```
acharb, line 64: @Ifropc is `postMessage` something that's used / we should support?

and if so what's an example of one? just a JSON string?

JakeUrban (commented):
We may want to hold this PR until we come to a conclusion on what changes we want to make to the SEPs regarding callbacks. 

There have been various discussions internally and with partners about whether the proposals should encourage callbacks to endpoints provided by the client on a per-transaction basis. It appears as if most businesses providing callback prefer to have client's register static endpoints outside of the transaction processing flow defined in SEPs. MoneyGram and Circle for example both take this approach to their respective callback services.

The two viable paths I see on this topic is to 

1. Keep the `callback` and `on_change_callback` parameters in the proposal, leaving it up to each anchor and wallet pair to determine whether and how they will implement support.
2. Deprecate and eventually remove the parameters and recommend that anchors require clients to register callback endpoints with them prior to initiating transactions.

src/walletSdk/Anchor/Sep24.ts
JakeUrban: We should not support `postMessage`. This is an error-prone way of signaling to the wallet to close the webview, and there are more straightforward approaches to this that have better user experiences as well.

Ifropc: 1. We are deprecating postMessage
2. We are also deprecating approach of passing callbacks "as is" to the request parameter. I'm working on drafting new approach, where callbacks are passed outside of the request.  
So let's hold on this PR for now

Ifropc:
Hey @acharb I think for now we can close this PR. We still need to discuss it more internally. Currently, in the Anchor Platform, which we use as a standard implementation of all SEPs, callbacks are configured on the Anchor side. So we may want to align protocol to the implementation, as we think this approach is better for various reasons

mtwtfss:
@Ifropc are you saying the Anchor Platform implementation currently deviates from the SEP standard and we think we might want to update the SEP standard to reflect this deviation?

any risk to merging what we have here for now while we figure out the changes we'd like to make?

also do you have any kind of ETA for when we'll land on a longterm solution?

Ifropc:
@mtwtfss 
Yes, that's correct!
There's no risk in merging it, but potentially we might carry deprecated parameters in the Wallet SDK, and would need to deal with the deprecation appropriately. 
Sorry for now there's no ETA, but I created a task for finalizing decisions of callback handling and put it on the top of my backlog

CassioMG:
@Ifropc do we have any news regarding the usage of sep24 callback parameters? I'm leaning towards closing this PR given it's around for a long time now

Ifropc:
I agree we can close this PR. It might be easier to re-do it in the future if necessary

PR #98, opened 2024-02-13
add KeyManager stuff from old wallet sdk

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1237?atlOrigin=eyJpIjoiODRmNDM3NGVjN2U0NDc4OGIwYWFhYWFjOGE5OWIyZGIiLCJwIjoiaiJ9)

largely just lifted from old wallets sdk


src/walletSdk/KeyManager/index.ts
```diff
@@ -0,0 +1,458 @@
+import { Networks, Transaction, WebAuth } from "@stellar/stellar-sdk";
+
+import { freighterHandler } from "./handlers/freighter";
+
+import { albedoHandler } from "./handlers/albedo";
+import { ledgerHandler } from "./handlers/ledger";
+import { plaintextKeyHandler } from "./handlers/plaintextKey";
+// TODO - fix trezor errors
+// import { trezorHandler } from "./handlers/trezor";
```
acharb, line 9: the trezor code is throwing some errors so I'm commenting out for a next PR

@piyalbasu for this error:
```
Cannot find module 'stellar-sdk' from 'node_modules/@trezor/connect-plugin-stellar/lib/index.js'

    Require stack:
      node_modules/@trezor/connect-plugin-stellar/lib/index.js
      src/walletSdk/KeyManager/handlers/trezor.ts
      src/walletSdk/KeyManager/index.ts
      src/index.ts
      test/keyManager.test.ts
```
Because trezor has a peer [dependency of stellar-sdk](https://github.com/trezor/trezor-suite/blob/62c48b2cf019528d31903dffe31fb8edf4f1a081/packages/connect-plugin-stellar/package.json#L27) , instead of `@stellar/stellar-sdk` 

That's something we need to ask them to change right? or do you know of a way to bypass that?

src/walletSdk/KeyManager/plugins/IdentityEncrypter.ts
```diff
@@ -0,0 +1,39 @@
...
+      }),
+      encrypterName: NAME,
+      salt: "identity",
+    });
+  },
+
+  decryptKey(params: DecryptParams) {
+    const { encryptedKey } = params;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { encrypterName, salt, encryptedBlob, ...props } =
```
acharb, line 31: @CassioMG the encrypterName and salt are being pulled out of props here so it's not returned below

do you know if that's on purpose? seems odd to me, since they're not being used

src/walletSdk/KeyManager/plugins/ScryptEncrypter.ts
```diff
@@ -0,0 +1,47 @@
...
+      encryptedBlob: encryptedPhrase,
+      encrypterName: NAME,
+      salt,
+    };
+  },
+
+  async decryptKey(params: DecryptParams) {
+    const { encryptedKey, password } = params;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { encrypterName, salt, encryptedBlob, ...props } = encryptedKey;
```
acharb, line 36: @CassioMG same thing here

acharb:
some dev dependencies w/ vulnerabilities looks like, looking into fixing those

acharb:
hmm looks like there's a dependency alert for [cross-fetch](https://socket.dev/npm/package/cross-fetch/overview/3.1.8), which is a dependency on trezor-connect

maybe another thing to ping trezor about, wdyt @piyalbasu ?

src/walletSdk/KeyManager/index.ts
CassioMG: Vibrant is only using `KeyType.plaintextKey` so we should be fine removing trezor unless Freighter is using it?

src/index.ts
```diff
@@ -27,6 +27,7 @@ export {
   TransactionBuilder,
   SponsoringBuilder,
 } from "./walletSdk/Horizon";
+export { KeyManager } from "./walletSdk/KeyManager";
```
CassioMG, line 30: I could be wrong here, but it's possible we'll need to add another line here to explicitly export the plugins as well. Something like:
`export * as KeyManagerPlugins from "./walletSdk/KeyManager/plugins";`

src/walletSdk/KeyManager/plugins/IdentityEncrypter.ts
CassioMG: It seems `encrypterName` and `salt` should be part of an `encrypted key` but not a `decrypted key`, so it's probably why they're being removed here to avoid adding extra unnecessary info to a `decrypted key`.

I've inferred it through `IdentityEncrypter's test` [here](https://github.com/stellar/js-stellar-wallets/blob/master/src/plugins/IdentityEncrypter.test.ts#L35), and we can also infer it from the return types of  `decryptKey` and  `encryptKey` declarations [here](https://github.com/stellar/js-stellar-wallets/blob/master/src/types/keys.ts#L84).

src/walletSdk/KeyManager/plugins/ScryptEncrypter.ts
CassioMG: Same answer from [here](https://github.com/stellar/typescript-wallet-sdk/pull/98#discussion_r1489626194).

We can also infer it through `ScryptEncrypter's test` [here](https://github.com/stellar/js-stellar-wallets/blob/master/src/plugins/ScryptEncrypter.test.ts#L32)

src/walletSdk/KeyManager/index.ts
acharb: cool, I'll leave it commented out for now

looks like freighter is only using plaintextKey too atm

src/walletSdk/KeyManager/plugins/IdentityEncrypter.ts
acharb: gotchya, makes sense thanks!

src/index.ts
acharb: ah good catch, thx!
https://github.com/stellar/typescript-wallet-sdk/pull/98/commits/30b7247cedd1e40eca6634ffbbc8d9f20ab7081b

piyalbasu:
> hmm looks like there's a dependency alert for [cross-fetch](https://socket.dev/npm/package/cross-fetch/overview/3.1.8) and [node-fetch](https://socket.dev/npm/package/node-fetch/overview/2.7.0), both from the trezor-connect package
> 
> maybe another thing to ping trezor about, wdyt @piyalbasu ?

Yeah, the `cross-fetch` has a medium supply chain security warning. I'm trying to understand what exactly caused that warning. But something worth flagging with Trezer

The `node-fetch` warning is less concerning. That warning is more about the fact that `node-fetch` can load remote code, which is an understood risk and reasonable for the Trezor lib to have to do

acharb:
closing [this PR](https://github.com/stellar/typescript-wallet-sdk/pull/102) in lieu of this one that's using yarn workspace

PR #99, opened 2024-02-20
WAL-1064 - add anchor platform integration tests

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1064?atlOrigin=eyJpIjoiNDg2NmY4N2JjNWIwNDFmNDllZGM1NjdlMWFjZTJiODEiLCJwIjoiaiJ9)


src/walletSdk/index.ts
```diff
@@ -73,16 +74,27 @@ export class Wallet {
...
    * things like getting the toml info.
+   * @param {allowHttp} [params.allowHttp] - Flag to allow http protocol for the home domain.
+   * Defaults to false, and can only be set to true on Testnet.
    * @param {string} [params.language=this.language] - The language setting for the Anchor.
    * @returns {Anchor} An Anchor instance.
    */
-  anchor({ homeDomain, language = this.language }: WalletAnchor): Anchor {
+  anchor({
+    homeDomain,
+    allowHttp = false,
```
acharb, line 84: allowing http for home domain anchors for testnet only

acharb:
or maybe _all_ tests? 🤔

src/walletSdk/Anchor/index.ts
```diff
@@ -19,6 +19,7 @@ import { parseToml } from "../Utils";
 type AnchorParams = {
   cfg: Config;
   homeDomain: string;
+  allowHttp?: boolean;
```
Ifropc, line 22: Wasn't allowHttp replaced with axios client configuration? Or do we still use allowHttp flag for that?
Probably should have it in our documentation 🤔

test/integration/anchorplatform.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      authToken,
+      params: {
+        asset_code: "USDC",
+        account: accountKp.publicKey,
+        type: "bank_account",
+        dest: "123",
+        dest_extra: "12345",
+      },
+    });
+    expect(wResp.id).toBeTruthy();
```
Ifropc, line 78: Can you also check transaction status?

test/integration/anchorplatform.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(transactionId).toBeTruthy();
+
+    const wResp = await anchor.sep24().withdraw({
+      withdrawalAccount: accountKp.publicKey,
+      assetCode,
+      authToken,
+    });
+    expect(wResp.id).toBeTruthy();
+
+    const transaction = await anchor.sep24().getTransactionBy({
```
Ifropc, line 100: Can you add a check for status ? 
Also, for withdrawal we can get transaction to pending_user_transfer_start

Ifropc: FYI: current Kotlin implementation automatically moves transaction between statuses, so I think we should include watcher in the test case

src/walletSdk/Anchor/index.ts
acharb: Yup the axios client can use http, but this is for something different. I needed to add because we're using the stellar-sdk Toml Resolver [here](https://github.com/stellar/js-stellar-sdk/blob/0e4a0b02d619dd1ef8c5c51e61c07f730c871581/src/stellartoml/index.ts#L34-L60), which is using their own axios client. And has a `allowHttp` config specific to that

ya good idea I'll add to docs

test/integration/anchorplatform.test.ts
acharb: this one actually doesn't have a status, the return body is just:
`{ id: 'd4c34a3d-4f2f-443c-b9be-320d385390b7' }`

test/integration/anchorplatform.test.ts
acharb: https://github.com/stellar/typescript-wallet-sdk/pull/99/commits/6251819e1cfe6d2582c3acb946ba8ca7d51c5258 👍 

for more full flow testing we already have tests in other files. If we want to do full anchor platform testing I think we can do when we move tests over, wdyt?

src/walletSdk/Anchor/index.ts
acharb: https://github.com/stellar/stellar-docs/pull/305

test/integration/anchorplatform.test.ts
Ifropc: No, I mean getting transaction and checking status

test/integration/anchorplatform.test.ts
Ifropc: Sounds good!

Ifropc (approved):

Issue #103, opened 2024-03-08
Error when trying to create account


yanielbf:
### What version are you using?
1.3.1

### What did you do?
 I am using React Native with Expo. I am triying to create account in Stellar Ledger

```
function generateRandomKeypair() {
  let wallet = new Wallet({
    stellarConfiguration: StellarConfiguration.TestNet(),
  });
  const account = wallet.stellar().account();
  const randomBytes = Crypto.getRandomBytes(32);
  return account.createKeypairFromRandom(Buffer.from(randomBytes));
}

async function handlerCreateAccount() {
    const parentAccount = SigningKeypair.fromSecret(
      "SBW5TWYWFRGPPCSBMLN2K7HDMKWUGLFPF5FMMSZQX7CT65EIQ5V5L43N"
    );
    const childAccount = generateRandomKeypair();

    let wallet = new Wallet({
      stellarConfiguration: StellarConfiguration.TestNet(),
    });
    const txBuilder = await wallet.stellar().transaction({
      sourceAddress: parentAccount,
    });
    const tx = txBuilder.createAccount(childAccount).build();
    parentAccount.sign(tx);
    await wallet.stellar().submitTransaction(tx);
}
```

I tested in codesanbox for react project and it's throwing same error

https://codesandbox.io/p/sandbox/loving-dewdney-s9n2dq?file=%2Fsrc%2FApp.js%3A25%2C29

### What did you expect to see?
Create new account in Stellar Ledger

### What did you see instead?
node_modules\@stellar\typescript-wallet-sdk\src\walletSdk\Horizon\Transaction\TransactionBuilder.ts

![image](https://github.com/stellar/typescript-wallet-sdk/assets/11427255/9bb63fd5-2885-4197-8477-b152a22174e0)

![Imagen de WhatsApp 2024-03-08 a las 01 18 09_4d14815a](https://github.com/stellar/typescript-wallet-sdk/assets/11427255/dae1e0bb-22a0-4d57-b0ab-1901f33e0479)

![image](https://github.com/stellar/typescript-wallet-sdk/assets/11427255/a327e21b-190f-494b-bbb5-7cd6220d814a)

acharb:
hey @yanielbf thanks for pointing this out. This was fixed in our 1.4.0 release which was just published. If you upgrade to that version it should be fixed

PR #104, opened 2024-03-08
Remove StellarSDK import and add Operation

yanielbf:
Remove the StellarSDK import because it is an undefined value and add the operation module in the import. This was causing this problem https://github.com/stellar/typescript-wallet-sdk/issues/103#issue-2175252006


acharb:
huh, good catch thanks @yanielbf 

It looks like importing the entire sdk comes up undefined in a React or React Native environment, but not in NodeJS. I left a ticket for the stellar-sdk team to take a look https://github.com/stellar/js-stellar-sdk/issues/931

I think we should remove all the StellarSDK imports to be safe

acharb:
changing them all here
https://github.com/stellar/typescript-wallet-sdk/pull/112

PR #105, opened 2024-03-13
[Soroban] Add getTokenInvocationArgs function

CassioMG:
https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37/backlog?selectedIssue=WAL-1365

### DIFF
--- a/src/walletSdk/Utils/soroban.ts
+++ b/src/walletSdk/Utils/soroban.ts
@@ -1,0 +1,84 @@
+import { Operation, StrKey, scValToNative, xdr } from "@stellar/stellar-sdk";
+
+import {
+  ArgsForTokenInvocation,
+  SorobanTokenInterface,
+  TokenInvocationArgs,
+} from "../Types";
+
+export const getArgsForTokenInvocation = (
+  fnName: string,
+  args: xdr.ScVal[],
+): ArgsForTokenInvocation => {
+  let amount: bigint;
+  let from = "";
+  let to = "";
+
+  switch (fnName) {
+    case SorobanTokenInterface.transfer:
+      from = StrKey.encodeEd25519PublicKey(
+        args[0].address().accountId().ed25519(),
+      );
+      to = StrKey.encodeEd25519PublicKey(
+        args[1].address().accountId().ed25519(),
+      );
+      amount = scValToNative(args[2]);
+      break;
+    case SorobanTokenInterface.mint:
+      to = StrKey.encodeEd25519PublicKey(
+        args[0].address().accountId().ed25519(),
+      );
+      amount = scValToNative(args[1]);
+      break;
+    default:
+      amount = BigInt(0);
+  }
+
+  return { from, to, amount };
+};
+
+export const getTokenInvocationArgs = (
+  hostFn: Operation.InvokeHostFunction,
+): TokenInvocationArgs | null => {
+  if (!hostFn?.func?.invokeContract) {
+    return null;
+  }
+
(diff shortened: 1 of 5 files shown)


src/walletSdk/Utils/soroban.ts
```diff
@@ -0,0 +1,84 @@
...
+      amount = scValToNative(args[2]);
+      break;
+    case SorobanTokenInterface.mint:
+      to = StrKey.encodeEd25519PublicKey(
+        args[0].address().accountId().ed25519(),
+      );
+      amount = scValToNative(args[1]);
+      break;
+    default:
+      amount = BigInt(0);
```
Ifropc, line 34: Should we throw here? It looks like this branch is inaccessible (we check that fnName is either transfer or mint before calling this function)

CassioMG: @Ifropc good catch! I'm in favor of `throwing` an error here, and also throwing in all places that are returning `null` so we make it clear to devs where is the flaw. What do you guys think?

@piyalbasu do we have a strong reason for returning `null` in all places [here](https://github.com/stellar/freighter/blob/master/extension/src/popup/helpers/soroban.ts#L143) including the `try catch` blocks instead of throwing?

piyalbasu: So our reasoning for returning values was to guard against changing xdr spec's. We basically wanted to be able to feed this methods different invocations and not have it break if a specific underlying key was unexpectedly named differently.

I think it's probably a better and clearer API to throw errors in a case where a user did not pass an appropriate Transfer or Mint invocation. We'll just need to let devs know that these methods expect a valid Transfer or Mint invocation, so they will need to check that before using these.

acharb:
hey @CassioMG PRs looking good thanks for handling!

Can we point this to the next release branch though instead of `main`? the branch is `release/1.4.0`

note: this branch changes the repo to a yarn workspace monorepo, so the project structure will be different (i dont think should cause any merge conflicts though since your code is in new files)

src/walletSdk/Utils/soroban.ts
CassioMG: @piyalbasu thanks for the follow-up!

Just to clarify, are you suggesting we only throw on the code block below?
```
  if (
    ![SorobanTokenInterface.transfer, SorobanTokenInterface.mint].includes(
      fnName,
    )
  ) {
    throw new Error("Please use a valid Transfer or Mint invocation.");
  }
```

CassioMG:
> hey @CassioMG PRs looking good thanks for handling!
> 
> Can we point this to the next release branch though instead of `main`? the branch is `release/1.4.0`
> 
> note: this branch changes the repo to a yarn workspace monorepo, so the project structure will be different (i dont think should cause any merge conflicts though since your code is in new files)

@acharb for sure! Sorry about the late response, I was OOO last week

CassioMG:
hey guys, thanks for the reviewing. I'm closing this PR in favor of https://github.com/stellar/typescript-wallet-sdk/pull/119/files which adds this same function in a separate `@stellar/typescript-wallet-sdk-soroban` workspace as suggested.

src/walletSdk/Utils/soroban.ts
CassioMG: Addressed this through a jsdoc comment here on the new PR: https://github.com/stellar/typescript-wallet-sdk/pull/119/files#diff-ebb0d904ab001286bca547ac642fe866bae6ffe07b009f97a6d8ad2d5aa5573eR40-R49

PR #111, opened 2024-03-19
add helper for parsing AnchorTransaction

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-1341?atlOrigin=eyJpIjoiZmQ1NDUwNTFjMjA4NDkwMTgyMWQ3MjM4N2Q1NWY5YjgiLCJwIjoiaiJ9)

### DIFF
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
@@ -23,6 +23,7 @@ import {
   ChallengeTxnIncorrectSequenceError,
   ChallengeTxnInvalidSignatureError,
   UnknownAnchorTransactionError,
+  InvalidJsonError,
 } from "../Exceptions";
 
 /**
@@ -75,7 +76,13 @@ export const signChallengeTransaction = async ({
 export const parseAnchorTransaction = (
   transaction: string,
 ): AnchorTransaction => {
-  const parsed = JSON.parse(transaction);
+  let parsed;
+  try {
+    parsed = JSON.parse(transaction);
+  } catch (e) {
+    throw new InvalidJsonError();
+  }
+
   if (parsed.kind === "withdrawal") {
     return parsed as WithdrawTransaction;
   } else if (parsed.kind === "deposit") {
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
@@ -76,17 +76,9 @@ export const parseAnchorTransaction = (
   transaction: string,
 ): AnchorTransaction => {
   const parsed = JSON.parse(transaction);
-  if (
-    "withdraw_memo_type" in parsed ||
-    "withdraw_memo" in parsed ||
-    "withdraw_anchor_account" in parsed
-  ) {
+  if (parsed.kind === "withdrawal") {
     return parsed as WithdrawTransaction;
-  } else if (
-    "deposit_memo" in parsed ||
-    "deposit_memo_type" in parsed ||
-    "claimable_balance_id" in parsed
-  ) {
+  } else if (parsed.kind === "deposit") {
     return parsed as DepositTransaction;
   } else if (parsed.status === "error") {
     return parsed as ErrorTransaction;
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
@@ -11,10 +11,18 @@ import {
 } from "@stellar/stellar-sdk";
 
 import { parseToml } from "../Utils";
-import { SignChallengeTxnParams, SignChallengeTxnResponse } from "../Types";
+import {
+  SignChallengeTxnParams,
+  SignChallengeTxnResponse,
+  AnchorTransaction,
+  WithdrawTransaction,
+  DepositTransaction,
+  ErrorTransaction,
+} from "../Types";
 import {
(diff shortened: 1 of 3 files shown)


@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
```diff
@@ -58,3 +66,31 @@ export const signChallengeTransaction = async ({
...
+ * Helper method for parsing a JSON string into an AnchorTransaction.
+ * @param {string} transaction - The json string of an anchor transaction.
+ * @returns {AnchorTransaction} The transaction object.
+ */
+export const parseAnchorTransaction = (
+  transaction: string,
+): AnchorTransaction => {
+  const parsed = JSON.parse(transaction);
+  if (
+    "withdraw_memo_type" in parsed ||
```
Ifropc, line 80: I think it's better to check `kind` instead (either withdrawal or deposit)

@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
```diff
@@ -58,3 +66,31 @@ export const signChallengeTransaction = async ({
...
+    "withdraw_memo" in parsed ||
+    "withdraw_anchor_account" in parsed
+  ) {
+    return parsed as WithdrawTransaction;
+  } else if (
+    "deposit_memo" in parsed ||
```
Ifropc, line 86: All of this could be null for incomplete transaction

@stellar/typescript-wallet-sdk/test/server.test.ts
```diff
@@ -57,3 +57,15 @@ describe("SEP-10 helpers", () => {
     expect(isValid).toBeFalsy();
   });
 });
+
+describe("Server helpers", () => {
+  it("should parse a JSON AnchorTransaction", () => {
```
Ifropc, line 62: Can you add a test for MGI incomplete transactions?

@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
acharb: that is way better thanks
https://github.com/stellar/typescript-wallet-sdk/pull/111/commits/4acb6305f75d0ae2f74bc44d31d091c5cdf7a09b

@stellar/typescript-wallet-sdk/test/server.test.ts
acharb: do you have a json string by chance? I'm having trouble generating one

Ifropc: I'll get one

piyalbasu (approved):
Looks good from a JS perspective. Once the incomplete tx test is in, we should be good

@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
```diff
@@ -58,3 +66,23 @@ export const signChallengeTransaction = async ({
...
+
+/**
+ * Helper method for parsing a JSON string into an AnchorTransaction.
+ * @param {string} transaction - The json string of an anchor transaction.
+ * @returns {AnchorTransaction} The transaction object.
+ */
+export const parseAnchorTransaction = (
+  transaction: string,
+): AnchorTransaction => {
+  const parsed = JSON.parse(transaction);
```
CassioMG, line 78: I believe `JSON.parse()` alone can throw, should we wrap it with `try catch` just in case we receive something not well formatted from the server?

CassioMG: And throw a different error in this case, something like `TransactionIsNotValidJSONError()` or similar.

Too conservative?

@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
```diff
@@ -58,3 +66,23 @@ export const signChallengeTransaction = async ({
...
+  if (parsed.kind === "withdrawal") {
```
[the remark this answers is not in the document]
acharb: hmm I like the idea but imo I think separating the parse functions would be better bc of the types. The types for sep6 and sep24 transactions are similar but have slight differences. I think if we'd want to add a parse method for sep6 it should be a separate method to make it clear what the method is returning 🤔  

right now it's returning `AnchorTransaction`, if it was returning `AnchorTransaction | Sep6Transaction` I think it would be a bit messy / too general

imo we can add in a separate PR if it's needed

if you feel strongly though lmk!

@stellar/typescript-wallet-sdk/test/server.test.ts
acharb: hey @Ifropc were you able to get one?

@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
acharb: ya I was on the fence if the try-catch should be handled here or by the function using this method.

But I think you've nudged me over the edge :)
https://github.com/stellar/typescript-wallet-sdk/pull/111/commits/4684622ba08d895371fce0535baabbe7d3d236e4

acharb:
spoke with @Ifropc  offline, going to merge this in to unblock it and will add the MGI test in a subsequent PR

@stellar/typescript-wallet-sdk/src/walletSdk/Server/index.ts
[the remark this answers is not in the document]
CassioMG: that makes sense to me!

PR #113, opened 2024-03-20
webpack fix

acharb:
when testing release/1.4.0 I realized one of the changes I made for the webpack build for react was causing an error for the nodeJS build, so this change further separates the environments

I added a [ticket](https://stellarorg.atlassian.net/browse/WAL-1404) to automate testing the different build environments, right now we're just auto testing nodeJS


piyalbasu (approved):

CassioMG (approved):
Nice catch!

PR #116, opened 2024-03-26
move husky to top level (interfering with publish)

acharb:


.github/workflows/npmPublishSdk.yml
```diff
@@ -15,6 +15,8 @@ jobs:
       - run: yarn test:ci
 
       - name: Publish to NPM
-        run: cd @stellar/typescript-wallet-sdk && yarn publish --access public
```
acharb, line 18: not changing anything, but it's better w/ pipe than `&&`

piyalbasu (approved):

.github/workflows/npmPublishSdk.yml
Ifropc: Out of curiosity -- why is it better?

PR #125, opened 2024-04-11
Use relative path to access "walletSdk" root folder

CassioMG:
For some reason accessing it with absolute `"walletSdk"` path is throwing `Cannot find module 'walletSdk' or its corresponding type declarations.` in several places while running `yarn example:sep24`.

Using a relative path like `"../"` fixes the issue.

Screenshot 2024-04-11 at 17 00 48


acharb (approved):
ya weird I wasn't seeing an error on mine, but nice fix

PR #162, opened 2024-08-21
add examples for sep-10

piyalbasu:


@stellar/typescript-wallet-sdk/examples/sep10/sep10Server.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable */
+
+// @ts-expect-error: express will need to be installed to run this example
```
piyalbasu, line 3: I thought it was maybe overkill to add `express` to the repo just for this example, so I just added this as a comment. Let me know if it's worth adding it so a user can actually spin this up from this repo

CassioMG: Is it too heavy? I'm personally fine adding it as a dev dependency as we could use it for other examples as well, but also OK in case you think that's too much

CassioMG (approved):
LGTM - left a few comments

aristidesstaffieri (approved):

@stellar/typescript-wallet-sdk/examples/sep10/sep10Server.ts
piyalbasu: I guess it won't affect the bundle size of the library, either. I'll add this as a dev dep 👍

@stellar/typescript-wallet-sdk/package.json
```diff
@@ -69,6 +70,7 @@
     "build:web": "webpack --config webpack.config.js",
     "build:node": "webpack --env NODE=true --config webpack.config.js",
     "build": "run-p build:web build:node",
+    "example:sep10": "ts-node examples/sep10/sep10Server.ts",
```
CassioMG, line 73: how should we run`sep10Wallet` ?

piyalbasu: Ah good call. I wasn't sure if this was useful to run as it's a simple helper, but I can add it!

PR #165, opened 2024-08-23
Add `transaction_id` field to SEP-12 requests

CassioMG:
Also improves typing by adding a separate `CustomerBinaryInfoMap` type only for the binary (Buffer) fields.

Closes: [Add transaction id to SEP-12 requests](https://github.com/orgs/stellar/projects/58/views/2?sliceBy%5Bvalue%5D=Iteration+6&pane=issue&itemId=73597661)

### DIFF
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Customer/index.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Customer/index.ts
@@ -2,13 +2,11 @@ import { AxiosInstance } from "axios";
 import queryString from "query-string";
 import { Sep9InfoRequiredError, CustomerNotFoundError } from "../Exceptions";
 import {
-  CustomerBinaryInfoMap,
   GetCustomerParams,
   GetCustomerResponse,
   AddCustomerResponse,
   AddCustomerParams,
   AuthToken,
-  CustomerInfoMap,
 } from "../Types";
 import { camelToSnakeCaseObject } from "../Utils";
 
@@ -98,7 +96,7 @@ export class Sep12 {
     memo,
     transactionId,
   }: AddCustomerParams): Promise<AddCustomerResponse> {
-    let customerMap: CustomerInfoMap | CustomerBinaryInfoMap = {
+    let customerMap: { [key: string]: string | Buffer } = {
       ...sep9Info,
       ...sep9BinaryInfo,
     };
@@ -151,7 +149,7 @@ export class Sep12 {
     memo,
     transactionId,
   }: AddCustomerParams): Promise<AddCustomerResponse> {
-    let customerMap: CustomerInfoMap | CustomerBinaryInfoMap = {};
+    let customerMap: { [key: string]: string | Buffer } = {};
     if (id) {
       customerMap["id"] = id;
     }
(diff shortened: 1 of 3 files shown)


aristidesstaffieri (approved):

@stellar/typescript-wallet-sdk/src/walletSdk/Customer/index.ts
```diff
@@ -78,28 +83,38 @@ export class Sep12 {
    * @param {AddCustomerParams} params - The parameters for adding a customer.
    * @param {CustomerInfoMap} [params.sep9Info] - Customer information. What fields you should
    * give is indicated by the anchor.
-   * @param {CustomerInfoMap} [params.sep9BinaryInfo] - Customer information that is in binary
+   * @param {CustomerBinaryInfoMap} [params.sep9BinaryInfo] - Customer information that is in binary
```
Ifropc, line 86: How does passing Buffer vs passing a string will work?
My question, is how exactly both of this types will be included in the request? 
Can we add a unit test that makes sure that the [body layout is correct](https://www.sobyte.net/post/2021-12/learn-about-http-multipart-form-data/#multipartform-data-request-body-layout) (I believe type of binary fields should NOT be `text/plain`)
And in the future add an integration test with Anchor Platform to pass [any of binary fields](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/api-schema/src/main/java/org/stellar/anchor/api/callback/PutCustomerRequest.java#L145-L166)? 

(I see we currently have tests with Polaris testanchor but it might just ignore this fields, so I don't think it's a very trustworthy test)

Ifropc (changes requested):

@stellar/typescript-wallet-sdk/src/walletSdk/Customer/index.ts
CassioMG: @Ifropc thanks for the review!
This is how it currently works: any `string` field should be passed through the `sep9Info` object and any `binary` field (e.g. KYC images) should be passed through the `sep9BinaryInfo` object.
If you pass any field on the `sep9BinaryInfo` object then the request [`Content-Type` header will be set to `multipart/form-data`](https://github.com/stellar/typescript-wallet-sdk/pull/165/files#diff-7e7a1e4a53167c9f357adaee4766bd01a15205d54eb38c528d2ac6097c578041R123).

How does that sound?

Ifropc: I was a bit confused by 
```
export type CustomerBinaryInfoMap = {
  [key: string]: Buffer | string;
};
```
My guess is that it will have an appropriate body layout depending on the type of the value in the map? 
In any case, adding a unit test would be a great addition.

Ifropc (approved):
LGTM, thanks for clarifying comment

@stellar/typescript-wallet-sdk/src/walletSdk/Customer/index.ts
CassioMG: > I was a bit confused by
> 
> ```
> export type CustomerBinaryInfoMap = {
>   [key: string]: Buffer | string;
> };
> ```

Fixed this is: https://github.com/stellar/typescript-wallet-sdk/pull/165/commits/6f38db4a08d5cdecdcf78f2a1400bd3b65a24362

Now the types look like this:
```
export type CustomerInfoMap = {
  [key: string]: string;
};

export type CustomerBinaryInfoMap = {
  [key: string]: Buffer;
};
```

> My guess is that it will have an appropriate body layout depending on the type of the value in the map?
> In any case, adding a unit test would be a great addition.

Created a task to tackle this unit test: https://github.com/orgs/stellar/projects/58/views/1?pane=issue&itemId=78332363

PR #166, opened 2024-08-24
Deprecate `getServicesInfo()` in favor of `Sep24.info()`

CassioMG:
We currently have the below `get info` functions for the related classes and SEPs:
- [Sep6.info()](https://github.com/stellar/typescript-wallet-sdk/blob/main/%40stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep6.ts#L65): hits `GET /info` for the `Sep6 server`
- [Sep38.info()](https://github.com/stellar/typescript-wallet-sdk/blob/main/%40stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep38.ts#L67): hits `GET /info` for the `Sep38 server`
- [Sep24.getServicesInfo()](https://github.com/stellar/typescript-wallet-sdk/blob/main/%40stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep24.ts#L183): hits `GET /info` for the `Sep24 server`
- [Anchor.getServicesInfo()](https://github.com/stellar/typescript-wallet-sdk/blob/main/%40stellar/typescript-wallet-sdk/src/walletSdk/Anchor/index.ts#L192): also hits `GET /info` for the `same Sep24 server` which could be `confusing`
 
The proposal of this PR is to `deprecate` both `getServicesInfo()` functions and consolidate them into a single `info()` function like the below for consistency and simplicity:
- `Sep6.info():` hits `GET /info` for the `Sep6 server`
- `Sep38.info()`: hits `GET /info` for the `Sep38 server`
- `Sep24.info()`: hits `GET /info` for the `Sep24 server`

This PR also adds support for the `lang` parameter on [Sep6.info()](https://github.com/stellar/typescript-wallet-sdk/pull/166/files#diff-6deabbb345fdb9b022d3f0daddf6fe94bb0030d34d389bf7f993cd7f9b38c663R69) and [Sep24.info()](https://github.com/stellar/typescript-wallet-sdk/pull/166/files#diff-166da6ffe01fd5f63003df55d66583420c883457a4edd9420f6a281384b8e11dR72) as covered by the [Sep-6](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#info) and [Sep-24](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#info) specs.

### DIFF
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep24.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep24.ts
@@ -15,6 +15,7 @@ import {
   GetTransactionsParams,
   AnchorServiceInfo,
   WatcherSepType,
+  Sep24Info,
 } from "../Types";
 import {
   Watcher,
@@ -44,6 +45,7 @@ export type Interactive = Sep24;
 export class Sep24 {
   private anchor: Anchor;
   private httpClient: AxiosInstance;
+  private anchorInfo: Sep24Info;
 
   /**
    * Creates a new instance of the Sep24 class.
@@ -55,6 +57,45 @@ export class Sep24 {
 
     this.anchor = anchor;
     this.httpClient = httpClient;
+  }
+
+  /**
+   * Get SEP-24 anchor information.
+   * If `shouldRefresh` is set to `true`, it fetches fresh values; otherwise, it returns cached values if available.
+   * @param {boolean} [shouldRefresh=false] - Flag to force a refresh of TOML values.
+   * @param {string} [lang=this.anchor.language] - The language in which to retrieve information.
+   * @returns {Promise<Sep24Info>} - SEP-24 information about the anchor.
+   * @throws {ServerRequestFailedError} If the server request to fetch information fails.
+   */
+  async info(
+    shouldRefresh?: boolean,
+    lang: string = this.anchor.language,
+  ): Promise<Sep24Info> {
+    if (this.anchorInfo && !shouldRefresh) {
+      return this.anchorInfo;
+    }
+
+    const { transferServerSep24 } = await this.anchor.sep1();
+    try {
+      const resp = await this.httpClient.get(
+        `${transferServerSep24}/info?lang=${lang}`,
+      );
+      this.anchorInfo = resp.data;
+      return resp.data;
+    } catch (e) {
+      throw new ServerRequestFailedError(e);
+    }
+  }
+
+  /**
+   * @deprecated Please use info() instead.
+   *
+   * Get SEP-24 anchor information.
+   * @returns {Promise<AnchorServiceInfo>} - SEP-24 information about the anchor.
+   * @throws {ServerRequestFailedError} If the server request to fetch information fails.
+   */
+  async getServicesInfo(): Promise<AnchorServiceInfo> {
+    return this.info();
   }
 
   /**
@@ -132,13 +173,13 @@ export class Sep24 {
     const toml = await this.anchor.sep1();
     const transferServerEndpoint = toml.transferServerSep24;
 
-    const serviceInfo = await this.anchor.getServicesInfo();
+    const sep24Info = await this.info();
 
     let assets: string[];
     if (type === FLOW_TYPE.DEPOSIT) {
-      assets = Object.keys(serviceInfo.deposit);
+      assets = Object.keys(sep24Info.deposit);
     } else {
-      assets = Object.keys(serviceInfo.withdraw);
+      assets = Object.keys(sep24Info.withdraw);
     }
     if (!assets.includes(assetCode)) {
       throw new AssetNotSupportedError(type, assetCode);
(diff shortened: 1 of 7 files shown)


@stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep24.ts
```diff
@@ -57,6 +59,45 @@ export class Sep24 {
...
+  /**
+   * Get SEP-24 anchor information.
+   * If `shouldRefresh` is set to `true`, it fetches fresh values; otherwise, it returns cached values if available.
+   * @param {boolean} [shouldRefresh=false] - Flag to force a refresh of TOML values.
+   * @param {string} [lang=this.anchor.language] - The language in which to retrieve information.
+   * @returns {Promise<Sep24Info>} - SEP-24 information about the anchor.
+   * @throws {ServerRequestFailedError} If the server request to fetch information fails.
+   */
+  async info(
+    shouldRefresh?: boolean,
```
aristidesstaffieri, line 71: Can the optional value come before the required one? Even though it has a default value, I don't think `shouldRefresh` can be in the first position and optional while `lang` is required.

CassioMG: @aristidesstaffieri thanks for the review!
I think both params are actually optional here, the `= this.anchor.language` initializer is there to set a default value for an optional param. Typescript won't let me do something like this: `lang?: string = this.anchor.language`.
The JSDoc also explicits that through the use of brackets syntax like `[paramName]` which is meant for optional params.

Since they both could be considered as optional we could invert them (like putting `lang` first), but the downside is that it would be a breaking change for people out there using the existing `Sep6.info()` function like this `Sep6.info(true)`, they would need to update it to `Sep6.info(undefined, true)`.

Does that make sense?

aristidesstaffieri: Got it yeah that does make sense. I guess the only downside is that users of `info` now can't leave off `shouldRefresh` while setting `lang` but this is probably fine imo.

aristidesstaffieri (approved):

@stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep24.ts
Ifropc: IMO lang should come first, as I don't see shouldRefresh being set very often (you generally should be fine with cached info)

Ifropc (approved):

@stellar/typescript-wallet-sdk/src/walletSdk/Anchor/Sep24.ts
CassioMG: I _think_ the `lang` param won't be used as much as well since in most cases it'll _probably_ be derived from `this.anchor.language`.

But I agree that putting _lang_ first does look better than having `shouldRefresh` first.

I guess we can revert them and make a BREAKING CHANGE note (which in reality I believe won't break for many people out there as it uses the default anchor language as mentioned earlier).

Another alternative yet is using an `object` as the `function input` so that the order of the parameters won't really matter. I think this object-input approach generally makes things more robust and scalable as it prevents bugs related to input params out of order and seems easier to maintain.

Something like this:
```
  async info({
    lang = this.anchor.language,
    shouldRefresh,
  }: {
    lang?: string,
    shouldRefresh?: boolean,
  }): Promise<Sep24Info> {
...
```

How do you guys feel about it?

aristidesstaffieri: I like the object argument approach and maybe we can handle the backwards compatability by switching on the logic of `typeof firstArg === boolean {...}`

Ifropc: > I think the lang param won't be used as much as well since in most cases it'll probably be derived from this.anchor.language.

That's a good point, I think you are 100% right

Ifropc: > Another alternative yet is using an object as the function input so that the order of the parameters won't really matter. I think this object-input approach generally makes things more robust and scalable as it prevents bugs related to input params out of order and seems easier to maintain.

+1 on that, but I think it will be breaking, won't it? Maybe a 2.0 change?

CassioMG: > Maybe a 2.0 change?
I like the 2.0 change approach. Maybe we could take this as an opportunity to review and make `all` SDK functions take objects as arguments. I see our current codebase has a mix of functions taking objects vs taking separate parameters, I think it would be great to establish this code convention on 2.0 and stick with it going forward.
What do you guys think?

We could merge this PR "as is" for `1.7` and plan for the wider change on `2.0`.

Ifropc: Sounds great!

CassioMG: Awesome, just created this issue to potentially tackle it in next iteration (sprint): [[TS Wallet SDK] Release 2.0 to consolidate object as function input](https://github.com/orgs/stellar/projects/58/views/1?pane=issue&itemId=78316509)

PR #237, opened 2026-05-21
Verify SEP-10 challenge body integrity after domain signing

CassioMG:
## Summary

- `Sep10.sign()` now verifies that the transaction returned by the configured `DomainSigner` has the same body hash as the original challenge before signing it with the user's account key. If the body differs, a `DomainSigningModifiedError` is thrown and the user-account signature is not applied.

## Test plan

- [x] Added unit test: *should accept when domain signer returns the same transaction with an appended signature* — confirms the legitimate flow still works.
- [x] Added unit test: *should reject when domain signer returns a transaction with a different body* — confirms a modified envelope causes `DomainSigningModifiedError` before any user-account signing.
- [x] `yarn lint` passes
- [x] `yarn test:ci` passes (249 tests, 0 failures)


piyalbasu (approved):
This is looking good to me!

PR #238, opened 2026-05-27
Correctness fixes: maxFee, 504 retries, key cache

CassioMG:
## Summary

Three independent correctness bug fixes, one commit each:

1. **`submitWithFeeIncrease` now respects `maxFee` on every retry**, not just the first — previously the fee cap was silently dropped after the first retry, letting fees grow unbounded.
2. **`submitTransaction` 504 retries are now bounded with exponential backoff** instead of recursing forever with no delay — avoids hammering Horizon.
3. **`KeyManager.removeKey` now clears the in-memory cache** — previously a removed key lingered in the cache and could still be used to sign when `shouldCache` was on.

No public API changes. No new dependencies.

---

## 1. Forward `maxFee` across `submitWithFeeIncrease` retries (`1dd8411`)

`submitWithFeeIncrease` rebuilt its params object on each recursive retry but omitted `maxFee`. As a result the cap was only enforced on the first retry; from the second retry onward `maxFee` was `undefined` and the guard `if (maxFee && …)` short-circuited to `false`, letting the fee grow unbounded.

- Forward `maxFee` in the recursive call.
- Tighten the guard from `maxFee && newFee > maxFee` to `maxFee !== undefined && newFee > maxFee`.

## 2. Bound `submitTransaction` 504 retries with exponential backoff (`c1dc76f`)

On HTTP 504 the method recursed with no retry limit, no delay, and no backoff. Under sustained Horizon 504s this hammered the already-slow server and could eventually exhaust the call stack (`RangeError: Maximum call stack size exceeded`), since V8 doesn't apply tail-call optimization to `await`.

- Convert the recursion to an iterative loop with a fixed five-retry budget.
- Capped exponential backoff: 1s base, 30s ceiling, additive jitter up to +50%.
- After exhausting retries, the last 504 is re-thrown.
- Use optional chaining on `e?.response?.status` to avoid a latent crash when a non-axios error enters the catch block.

## 3. Clear `KeyManager` cache entry on `removeKey` (`e2d9691`)

`_writeIndexCache`'s guard `if (this.shouldCache && key)` skipped the body when `key` was `undefined`, so `removeKey`'s `_writeIndexCache(id, undefined)` call was a no-op. With `shouldCache: true`, a removed key stayed in the in-memory cache and subsequent `signTransaction` / `fetchAuthToken` calls kept returning it.

- Split the guard so the method writes when given a key and deletes the cache entry when given `undefined`.

## Testing

Each fix ships with a focused unit test:

- `submitWithFeeIncrease` cap is enforced across multiple retries (not just the first).
- `submitTransaction` 504 handling: retries then succeeds, re-throws non-504 immediately, and gives up after the retry budget.
- `removeKey` purges the cache so a subsequent sign attempt misses and fails as expected.

No public API changes. No new dependencies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
@@ -152,12 +152,17 @@ export class Stellar {
         }
         // https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/timeout
         // https://developers.stellar.org/docs/encyclopedia/error-handling#timeouts
+        // Equal-jitter backoff: each attempt waits at least half the capped
+        // exponential delay (a deterministic, progressive floor) plus a random
+        // amount up to the other half. Keeps the schedule predictable while
+        // smoothing correlated retries across many clients during a Horizon
+        // outage. Total sleep stays within SUBMIT_504_MAX_DELAY_MS.
         const cappedDelay = Math.min(
           SUBMIT_504_BASE_DELAY_MS * 2 ** attempt,
           SUBMIT_504_MAX_DELAY_MS,
         );
-        const jitter = Math.random() * (cappedDelay / 2);
-        await sleep(cappedDelay + jitter);
+        const half = cappedDelay / 2;
+        await sleep(half + Math.random() * half);
       }
     }
     throw lastError;
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
@@ -199,7 +199,7 @@ export class Stellar {
       if (resultCode === "tx_too_late") {
         const newFee = parseInt(transaction.fee) + baseFeeIncrease;
 
-        if (maxFee && newFee > maxFee) {
+        if (maxFee !== undefined && newFee > maxFee) {
           throw new TransactionSubmitWithFeeIncreaseFailedError(maxFee, e);
         }
 
@@ -211,6 +211,7 @@ export class Stellar {
           signerFunction,
           baseFee: newFee,
           memo,
+          maxFee,
         });
       }
       throw e;
--- a/@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
+++ b/@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
@@ -24,6 +24,13 @@ import {
 import { getResultCode } from "../Utils/getResultCode";
 import { SigningKeypair } from "./Account";
 
+const SUBMIT_504_MAX_RETRIES = 5;
+const SUBMIT_504_BASE_DELAY_MS = 1000;
+const SUBMIT_504_MAX_DELAY_MS = 30000;
+
+const sleep = (ms: number) =>
+  new Promise<void>((resolve) => setTimeout(resolve, ms));
+
 /**
  * Interaction with the Stellar Network.
  * Do not create this object directly, use the Wallet class.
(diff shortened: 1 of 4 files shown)


@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
```diff
@@ -199,7 +218,7 @@ export class Stellar {
       if (resultCode === "tx_too_late") {
         const newFee = parseInt(transaction.fee) + baseFeeIncrease;
 
-        if (maxFee && newFee > maxFee) {
+        if (maxFee !== undefined && newFee > maxFee) {
```
piyalbasu: **`maxFee: 0` regression.** This guard changed from `if (maxFee && newFee > maxFee)` to `if (maxFee !== undefined && newFee > maxFee)`. The old behavior treated `maxFee === 0` as falsy ("no cap"); the new behavior treats it as a hard zero cap that always trips, since `newFee = parseInt(transaction.fee) + baseFeeIncrease > 0`.

Any caller passing `maxFee: 0` (defensive default, numeric coercion of an unset env var, or as a documented no-cap sentinel under the old behavior) will now throw `TransactionSubmitWithFeeIncreaseFailedError(0, e)` on the very first `tx_too_late` and never resubmit. Worth a CHANGELOG note flagging the breaking change, or — if 0 was previously undocumented — at least narrowing the type to a positive number.

Separately, this is also bypassed when `parseInt(transaction.fee)` returns `NaN`: `NaN > maxFee` is false, so `submitWithFeeIncrease` recurses with `baseFee: NaN`. Worth guarding with `Number.isFinite(newFee)`.

@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
```diff
@@ -127,21 +134,33 @@ export class Stellar {
...
-      const response = await this.server.submitTransaction(signedTransaction);
-      if (!response.successful) {
-        throw new TransactionSubmitFailedError(response);
-      }
-      return true;
-    } catch (e) {
-      if (e.response.status === 504) {
-        // in case of 504, keep retrying this tx until submission succeeds or we get a different error
+    let lastError: unknown;
+    for (let attempt = 0; attempt <= SUBMIT_504_MAX_RETRIES; attempt++) {
```
piyalbasu, line 149: **504 retry semantics changed from unbounded to ~6 attempts / ~46s.** The previous implementation retried recursively forever; this caps at `SUBMIT_504_MAX_RETRIES + 1` attempts and then throws the last 504. Two follow-ups worth considering:

1. **Caller cascade**: `submitWithFeeIncrease` catches `submitTransaction` errors and only handles `tx_too_late`. A bubbled 504 has `getResultCode(e) === ''`, so it rethrows immediately — even though the original transaction may still land on-chain. A naive caller that retries on its own will double-submit. Worth either documenting this in the JSDoc or special-casing 504 in `submitWithFeeIncrease`.

2. **Jitter exceeds the documented cap** (line 160): `cappedDelay = Math.min(BASE * 2**attempt, MAX)` is clamped to 30s, but jitter (`Math.random() * cappedDelay/2`, up to 15s) is then *added on top*, giving ~45s waits on the final retries — 50% over `SUBMIT_504_MAX_DELAY_MS`. If the cap is meant to bound caller latency, apply jitter first and `Math.min` after, or use full-jitter (`Math.random() * cappedDelay`).

piyalbasu (commented):
A couple of correctness concerns on the new `submitTransaction` / `submitWithFeeIncrease` logic.

piyalbasu (approved):
This looks good to me. I'd say the `maxFee` comment is worth addressing, but the 504 improvement could maybe be pushed off

@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
CassioMG: On `maxFee: 0`: agreed, reverted in 4939bb2 — the guard is back to `if (maxFee && newFee > maxFee)`, preserving the `maxFee: 0 ≈ no cap` behavior. The actual bug fix (forwarding `maxFee` across the recursive call) is what closes the report; the guard tightening was spec creep.

On `NaN` guard: pushing back. `transaction.fee` comes from `@stellar/stellar-sdk`'s `TransactionBuilder`, which sets `fee` from a numeric value — it's always a numeric string. For `parseInt` to return `NaN` here, the SDK contract would have to break, and we already depend on that contract elsewhere (`baseFee.toString()` in `makeFeeBump`, the builder constructor, etc.). I'd rather not sprinkle `Number.isFinite` defenses against contracts we already lean on; happy to revisit if there's a concrete path where `transaction.fee` could be non-numeric.

@stellar/typescript-wallet-sdk/src/walletSdk/Horizon/Stellar.ts
CassioMG: On 1 (caller cascade): you're right that an exhausted-504 bubbles through `submitWithFeeIncrease` unchanged, and the developer-experience concern stands. Documented in e7aba81 — the new JSDoc on `submitTransaction` explicitly calls out that an exhausted-504 leaves the on-chain status indeterminate and that callers should poll the tx hash rather than resubmit blindly (and notes that resubmitting the same signed tx will fail with `tx_bad_seq` once the original lands, so it's a DX problem rather than a fund-loss one — Stellar's sequence numbers prevent the double-submit you're concerned about).

On 2 (jitter exceeds the cap): fixed in ba3cb75 — switched to equal-jitter (`cappedDelay/2 + Math.random() * cappedDelay/2`), so sleeps now stay in `[cap/2, cap]`. Retains the herd-smoothing benefit with a deterministic progressive floor, and the named `SUBMIT_504_MAX_DELAY_MS` actually bounds the wait now.

CassioMG:
> This looks good to me. I'd say the `maxFee` comment is worth addressing, but the 504 improvement could maybe be pushed off

@piyalbasu Both your comments have been addressed, thanks for the review!

PR #140, opened 2024-05-31
Add new on_hold status support

Ifropc:
New status was introduced in SEP-24 protocol change in https://github.com/stellar/stellar-protocol/pull/1479/
This PR adds new status

PR #251, opened 2026-08-27
fix(watcher): treat on_hold as transient in watchOneTransaction

yigitcangokmen:
## Problem

`watchOneTransaction` classifies `on_hold` as a terminal error. The status falls through to the final `else`, which calls `onError(transaction)` and schedules no further poll, so the watch stops.

Per SEP-24, `on_hold` is transient: a deposit or withdrawal is placed on hold for additional checks after the user's funds are received, and once cleared it follows the regular flow. The SDK's own JSDoc on the enum member says the same, and the `watchOneTransaction` JSDoc enumerates the intended error statuses (`no_market`, `too_small`, `too_large`, `error`), which do not include `on_hold`. So this is an oversight, not intended behavior.

`on_hold` was added to `TransactionStatus` in #140 (2024-06-28), which changed only the types file; the `Watcher` was never updated for it.

## Impact

A wallet tracking a single transfer with `Sep24.watcher()` or `Sep6.watcher()` (both use `watchOneTransaction`) reports a false failure when an anchor puts a legitimate transfer on hold for compliance review, and the eventual `completed` or `refunded` is never delivered.

## Fix

One line: treat `on_hold` as in-progress (keep polling, emit `onMessage`), consistent with how `watchAllTransactions` handles it, which never routes `on_hold` to `onError` and never stops polling on it.

## Tests

Added an `on_hold` case to the `watchOneTransaction` suite: a watch going `pending_anchor -> on_hold -> completed` fires no `onError`, keeps polling, and observes `completed`. The test fails on `main` (the watch errors out at `on_hold`) and passes with this change; all existing watcher tests still pass.

## Scope

- Reproduced and verified on the SEP-24 path; SEP-6 is affected through the same shared `Watcher.watchOneTransaction` code path.
- `watchAllTransactions` already keeps polling on `on_hold`; its first-load handling of an already-`on_hold` transaction is separate behavior and is out of scope here.


CassioMG:
@yigitcangokmen Thanks for picking this up — well-scoped fix, and the regression test makes the behavior change easy to verify. Great to see a contribution to the SDK from outside the core team.

One request before this merges (plus a small docs nit in a separate comment below).

### `watchAllTransactions` has the same `on_hold` gap — worth closing here too

**Suggestion / consistency** — Not a merge blocker for the one-line fix as written, but we'd like it addressed in this PR rather than left as a follow-up.

**TL;DR:** The sibling watcher method has the same blind spot this PR just fixed: it doesn't recognize `on_hold` as an in-progress status. The effect there is different, and arguably worse — instead of reporting a false error, a transfer that is already on hold the first time the watcher sees it is silently dropped and never reported again, even after it completes. Since you've already established that `on_hold` is transient, it would be good to close both halves in one change so the two watchers agree.

**Steps to reproduce:**
1. Start a watcher over all transactions for an asset, without passing the transaction's id in the `watchlist`.
2. Have the anchor return a transaction that is already `on_hold` on the very first poll.
3. Let the anchor move that transaction to `completed` on a later poll.
4. Neither the on-hold state nor the eventual completion is ever delivered to any callback.

---

<details>
<summary>Detailed explanation (for agents)</summary>

**Root cause:** `watchAllTransactions` computes its own `isInProgress` predicate, which was never updated when `on_hold` was added to `TransactionStatus`:

https://github.com/stellar/typescript-wallet-sdk/blob/cbdb03b04d3934725a87618674174012dd495922/@stellar/typescript-wallet-sdk/src/walletSdk/Watcher/index.ts#L168-L172

On the first poll (`isRetry === false`), any transaction that is not in progress and not on the caller's watchlist is written into `_transactionsIgnoredRegistry` and filtered out:

https://github.com/stellar/typescript-wallet-sdk/blob/cbdb03b04d3934725a87618674174012dd495922/@stellar/typescript-wallet-sdk/src/walletSdk/Watcher/index.ts#L182-L189

That registry entry is permanent for the life of the watcher. The later branch that rescues a transaction which reached a final status is explicitly gated on the id *not* being in the ignored registry:

https://github.com/stellar/typescript-wallet-sdk/blob/cbdb03b04d3934725a87618674174012dd495922/@stellar/typescript-wallet-sdk/src/walletSdk/Watcher/index.ts#L198-L211

So an already-`on_hold` transaction first observed on the initial poll reaches neither `onMessage` nor `onError`, ever — including once it becomes `completed` / `refunded` / `expired` / `error`.

For what it's worth, the PR description's claim that `watchAllTransactions` "never routes `on_hold` to `onError` and never stops polling on it" is accurate as far as it goes: a transaction *already tracked* from an earlier `pending_*` state does transition through `on_hold` correctly, because the `registeredTransaction.status !== transaction.status` fallback masks the gap on that path. It is specifically the first-load path that breaks.

**Suggested fix** — the same one-line change you already made, mirrored:

```ts
const isInProgress =
  transaction.status.indexOf("pending") === 0 ||
  transaction.status === TransactionStatus.incomplete ||
  transaction.status === TransactionStatus.on_hold;
```

A regression test in the `watchAllTransactions` suite in `@stellar/typescript-wallet-sdk/test/wallet.test.ts`, shaped like the one you added, would cover it: first poll returns an already-`on_hold` transaction that is not on the watchlist, second poll returns it `completed`; assert it is delivered via `onMessage` on both polls and that `onError` never fires.

**Context:** `on_hold` was introduced in #140, which touched only `Types/anchor.ts` — neither watcher was updated at the time, which is why both carry the same gap.

</details>

CassioMG:
### `watchOneTransaction` JSDoc no longer matches the callback contract

**Nit / docs** — Not a merge blocker.

**TL;DR:** The doc block above the method spells out which statuses land in which callback, and `on_hold` now belongs in the "keep watching" bucket but isn't listed there. That block is the only place this callback contract is documented, so it's worth updating in the same change. (Copilot [flagged this one](https://github.com/stellar/typescript-wallet-sdk/pull/251#discussion_r3872406771) as well.)

---

<details>
<summary>Detailed explanation (for agents)</summary>

The JSDoc enumerates the statuses per callback:

https://github.com/stellar/typescript-wallet-sdk/blob/cbdb03b04d3934725a87618674174012dd495922/@stellar/typescript-wallet-sdk/src/walletSdk/Watcher/index.ts#L276-L282

Line 278 still reads `onMessage - When the transaction comes back as pending_ or incomplete.`, but after this change `on_hold` also routes to `onMessage`:

https://github.com/stellar/typescript-wallet-sdk/blob/cbdb03b04d3934725a87618674174012dd495922/@stellar/typescript-wallet-sdk/src/walletSdk/Watcher/index.ts#L368-L374

**Suggested fix:**

```
* * onMessage - When the transaction comes back as pending_, incomplete or on_hold.
```

The PR description already leans on the `onError` line of this same doc block as evidence that `on_hold` was never meant to be terminal — a good argument; this just completes the other half of it.

</details>

yigitcangokmen:
Done @CassioMG  🦆🦆🦆 
`watchAllTransactions` now treats `on_hold` as in-progress too, and I updated the     
  `watchOneTransaction` JSDoc to list it.
                                                                                                    
  Two tests added for the `watchAllTransactions` side:                                              
                                                                                                    
  - **silent-drop scenario** : an `on_hold` transfer on the very first poll is reported via         
  `onMessage` instead of being silently ignored, and the `completed` transition shows up as a status
   change                                                                                           
  - **duplicate guard** : when the same `on_hold` status comes back on consecutive polls, it's only
  reported once

CassioMG (approved):
@yigitcangokmen this LGTM! Thanks for the contribution.

